PR #122, opened 2014-02-07
Add support for has_child / has_parent filter

carloslopes:
I didn't find a pratice way to test this.
Maybe someone can help. :)

Example usage:
```ruby
Product.search('*', has_child: { type: <child document type>, where: <hash of where filters> }, has_parent: { type: <parent document type>, where: <hash of where filters> })
```

### DIFF
--- a/lib/searchkick/search.rb
+++ b/lib/searchkick/search.rb
@@ -267,6 +267,19 @@ module Searchkick
 
       # filters
       filters = where_filters.call(options[:where])
+
+      # has_child filter
+      if value = options[:has_child]
+        filters << {
+          has_child: {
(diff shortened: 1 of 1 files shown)


khelal:
+1

carloslopes:
I'm closing this PR for two reasons.

1. It was made using the old searchkick structure, and now everything is different inside it;
2. The change made here, solves only the search problem. And this would cause problems when we reindex the documents or do other changes that involve the parent/child documents. So, to support this feature completely, it's necessary to cover other aspects and this require a lot of work and planning to not break the gem;

If someone wants to help me to plan and implement this, it would be very nice.

PR #160, opened 2014-03-26
added the possibility to use a proc or lambda as index_name

jdurand:
This was the only thing missing to make searchkick compatible with my multi-tenanted app.
Using [Apartment](https://github.com/influitive/apartment), I can just do this : 

``` ruby
searchkick index_name: -> { "#{self.name.tableize}-#{Apartment::Database.current_tenant}" }
```

...and voilà!

### DIFF
--- a/lib/searchkick/model.rb
+++ b/lib/searchkick/model.rb
@@ -3,17 +3,19 @@ module Searchkick
 
     def searchkick(options = {})
       class_eval do
-        cattr_reader :searchkick_options, :searchkick_env, :searchkick_klass, :searchkick_index
+        cattr_reader :searchkick_options, :searchkick_env, :searchkick_klass
 
         class_variable_set :@@searchkick_options, options.dup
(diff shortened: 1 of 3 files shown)


sandeepravi:
Is there any way a single "tenant" can be reindexed?

jdurand:
Lets say you use the Apartment gem:

```ruby
tenant = "example.com"
Apartment::Tenant.switch(tenant)
  I18n.available_locales.each do |locale|
    I18n.locale = locale
    puts "Reindexing Things for #{tenant}-#{locale}..."
    Thing.reindex
end
```

sandeepravi:
I think I should have explained myself a little better.

Let's say I have a Thing model - There can be different keys in the "Thing" model (Mongodb) for each tenant. This is not much of an issue as ES should be able to handle it out of the box. But the "Thing" model can have different analyzers as well. One tenant can have a location key while the other need not. One tenant can have autocomplete on title while the other need not. While this can be done, I want to be able to run,

`Doc.reindex(import:` false) 

for one particular tenant so that the mapping changes only for that tenant.

Issue #239, opened 2014-07-16
Highlighting of text_middle matches


Dannyguk:
Hi

Firstly thanks for the excellent Searchkick :)

I'm having issues getting highlighting when results are matched using text_middle. For example I have an index with URLs in it, the highlight tag is included for all matches except the URL itself.

example of the highlight array below for the query "travel", using the highlight tag "<mark>".

{:highlight=>{:"url.text_middle"=>"http://www.wmtravellers.com/", :bio=>"We love to <mark>travel</mark> - wildlife, road trips, gorgeous scenery & photography are our passions!!", :campaign_tag=>"eco, ethical, <mark>travel</mark>, wildlife, environment", :location=>"<mark>Travelling</mark>"}}

Many Thanks
Dan

ankane:
Cleaning up issues - Searchkick now has a `highlight` option.

PR #242, opened 2014-07-24
add text middle tag test #239

Dannyguk:
Hi 

As discussed in issue #239.

Thanks
Dan


ankane:
Cleaning up pull requests.

PR #283, opened 2014-09-18
add indices_boost option support

exAspArk:
Hi, I'd like to add [indices_boost](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/search-request-index-boost.html) support. For example:

```ruby
User.search("*", index_name: ["foo", "bar"], indices_boost: { foo: 1, bar: 2 })
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,12 @@ boost_where: {user_id: {value: 1, factor: 100}} # default factor is 1000
 boost_where: {user_id: [{value: 1, factor: 100}, {value: 2, factor: 200}]}
 ```
 
+Boost indices when searching across more than one index
+
+```ruby
+indices_boost: { Animal => 1, Product => 200 }
+```
+
 [Conversions](#keep-getting-better) are also a great way to boost.
 
 ### Get Everything
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -345,6 +345,9 @@ module Searchkick
         # order
         set_order(payload) if options[:order]
 
+        # indices_boost
+        set_boost_by_indices(payload)
+
         # filters
         filters = where_filters(options[:where])
         set_filters(payload, filters) if filters.any?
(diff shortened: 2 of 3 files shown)


jslauthor:
It's a shame this was never accepted. I had to write a custom query to get around it. :/

ankane:
Hey @exAspArk, sorry for letting this stagnate.  It looks like there's still an issue with the indices_boost and aliases.  I'm not excited about the performance implications of this approach (an additional Elasticsearch call for each index), but seems like the best we can do.  We could cache the aliases, but would need to invalidate them after a reindex, which could be hard to detect without additional complexity.  I've added a few comments inline.  Happy to merge once they are addressed.

README.md
```diff
@@ -138,6 +138,12 @@ boost_where: {user_id: 1} # default factor is 1000
 boost_where: {user_id: {value: 1, factor: 100}}
 ```
 
+Boost indices when searching across more than one index
+
+```ruby
+indices_boost: { Animal.searchkick_index.name => 1, Product.searchkick_index.name => 200 }
```
ankane: Can we also allow `indices_boost: {Animal: 1, Product: 200}`? We do this with the `index_name` option.

searchkick.gemspec
```diff
@@ -25,4 +25,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "bundler", "~> 1.6"
   spec.add_development_dependency "rake"
   spec.add_development_dependency "minitest"
+  spec.add_development_dependency "pry"
```
ankane: I'd prefer to keep development dependencies to a minimum.

test/query_test.rb
```diff
@@ -11,4 +11,21 @@ def test_basic
     assert_equal ["Apple", "Milk"], query.execute.map(&:name).sort
   end
 
+  def test_indices_boost
+    store_names(["Rex"], Animal)
+    store_names(["Rexx"], Product)
+    index_names = [Animal.searchkick_index.name, Product.searchkick_index.name]
+    indices_boost = { Animal.searchkick_index.name => 1, Product.searchkick_index.name => 200 }
+
+    query = Store.search("Rex", index_name: index_names, indices_boost: indices_boost, execute: false)
```
ankane: I think we can simplify the test significant. Rather than check the generated body, let's check the order of the results with `assert_order`.

exAspArk:
Hey @ankane, thanks for your feedback! I've pushed the changes.

Yep, it's sad that we have to make additional requests while the issue is still open https://github.com/elastic/elasticsearch/issues/4756.

README.md
```diff
@@ -168,6 +168,12 @@ boost_where: {user_id: {value: 1, factor: 100}} # default factor is 1000
 boost_where: {user_id: [{value: 1, factor: 100}, {value: 2, factor: 200}]}
 ```
 
+Boost indices when searching across more than one index
```
ankane: Can we move this to a section about searching multiple models? (maybe right above the reference section)

ankane: Actually, I can take this one. Thanks @exAspArk!

ankane:
:+1:

PR #289, opened 2014-09-25
Allow for primary_key to be passed as an option

jimiray:
We had a need in an project to be able to look up records from an alternate primary key, so I added it as an option to be passed in.


ankane:
Cleaning up pull requests.

PR #290, opened 2014-09-25
added id transformation for arrays in sorting

Asp1d:
This fix allows to use multiple sorting with directions


ankane:
Cleaning up pull requests.  Please re-open if there's another test case this covers.

PR #296, opened 2014-10-02
Adds query function parameter to boost_where

anfinil:
In some cases it is actual to have other function_score query than the `term` query. The pull requests add such ability.


anfinil:
Is any problem?

PR #321, opened 2014-11-12
Rescue from division by zero error

sharshenov:
`per_page` can be 0


ankane:
Thanks @sharshenov!  Kaminari raises an error in this situation.  Searchkick should probably do the same. https://github.com/amatsuda/kaminari/blob/master/lib/kaminari/models/page_scope_methods.rb#L48

ankane:
Cleaning up PRs

PR #331, opened 2014-11-18
Update README.md

macarthy:
Hope you don't think this is a stupid point! I changed the the "2% Milk" to "Low fat Milk" for clarity. An improvement because, Non-USA programmers are not familiar with "2% Milk" and also "2%" looks like a programming term, or placeholder in many languages, printf statements etc. A pure string is easier to parse in docs.


ankane:
Sorry for the delayed merge.  Good point - changed to `apples` instead to avoid any confusion.

PR #355, opened 2015-01-05
Added support for geo_shape with readme and tests

DmitryKK:
Sorry, if there are grammatical errors. My English is bad(

### DIFF
--- a/README.md
+++ b/README.md
@@ -657,7 +657,13 @@ class City < ActiveRecord::Base
 end
 ```
 
-Your coordinates can be presented either as a string (searchkick will convert them into array) or as an array.
+Your coordinates can be presented either as a string (searchkick will convert them into array) or as an array. For example:
+
+```ruby
+  ...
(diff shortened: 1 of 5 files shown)


ankane:
Thanks @DmitryKK and @sirsupertramp for #361. Sorry for letting these PRs stagnate. I'd like to add geo shape support pulling the best from both of your approaches.  Here's what I propose:

```ruby
class Product
  searchkick geo_shape: {
    boundingbox: {tree: "quadtree", precision: "1km"}
  }

  def search_data
    {
      boundingbox: {type: "envelope", coordinates: [{lat: 1, lon: 2}, {lat: 3, lon: 4}, {lat: 5, lon: 6}]}
    }
  end
end
```

And search with:

```ruby
Product.search("*", where: {boundingbox: {geo_shape: {type: "circle", coordinates: [{lat: 2, lon: 3}], radius: "100m"}}}})
```

Thoughts?  We could also allow for coordinates to be specified as any array, but Searchkick switched to `lat` and `lon` for other features to prevent confusion about the order with an array.

ankane:
Created a help wanted request here #765

PR #358, opened 2015-01-08
Refactor ReindexJob

maxbeizer:
@ankane noticed this today and took a stab at it.

Now that the `perform` methods are identical, I would love to extract them to a common method and include them in both reindex job classes.  What do you think? :beers:

### DIFF
--- a/lib/searchkick/reindex_job.rb
+++ b/lib/searchkick/reindex_job.rb
@@ -1,5 +1,6 @@
 module Searchkick
   class ReindexJob
+    attr_reader :klass, :id
 
     def initialize(klass, id)
       @klass = klass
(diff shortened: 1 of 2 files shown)


maxbeizer:
I'm seeing intermittent failures for three tests in `test/suggest_test.rb`. I'm not sure how this PR and those test are related yet. I'll look into it further.

maxbeizer:
These failures can be reproduced with `rake TESTOPTS="--seed=44864"`

I'm still working on a solution.

maxbeizer:
I believe I have found the solution, but CI is still failing. I think this is because I fixed up my commit and CI is still building commit `f51fdd3` instead of `f6fabb1`

ankane:
Thanks @maxbeizer!  The original reindex job will most likely be removed when Searchkick 1.0 is released, so think it's best to hold off on a refactor.

maxbeizer:
No worries, @ankane . Thanks for this gem. :beers:

PR #361, opened 2015-01-10
Added geoshape (boundingbox) query support.

sirsupertramp:
Added query to search for shapes (boundingboxes) that imbed a specified
point (location). Also added reference to README.


ankane:
Created a help wanted request here #765

PR #368, opened 2015-01-21
Allow '_all' to be used as a search field

nornagon:
This lets you set index type (e.g. `word_middle`) for some fields but still search every field in the index.


ankane:
Thanks for the pull request @nornagon!  Can you add tests and update the readme?

ankane:
Cleaning up PRs. Feel free to reopen with tests.

Issue #187, opened 2014-04-19
routing


plasticgit:
is there a way to take advantage of custom document routing on searches as outlined here http://www.elasticsearch.org/blog/customizing-your-document-routing/ ?

bogdanovich:
+1
This video describes how routing could increase performance in user data flow
http://www.elasticsearch.org/videos/big-data-search-and-analytics/

ankane:
Now possible thanks to @bogdanovich

PR #369, opened 2015-01-22
Add routing parameter

bogdanovich:
This commit adds routing parameter so we could search this way:

```ruby
Contact.search "Joe", routing: "12"
```

In order to get proper search results with routing you should reindex with following mapping:
```ruby
searchkick merge_mappings: true, mappings: {
  contact: { _routing: { required: true, path: "user_id" } }
}
```

Routing increases performance when you have user data flow.

Designing for scale article (see user data flow):
http://www.elasticsearch.org/guide/en/elasticsearch/guide/current/user-based.html
http://www.elasticsearch.org/guide/en/elasticsearch/guide/current/routing-value.html

Data design patterns (video): 
http://www.elasticsearch.org/videos/big-data-search-and-analytics/ 

Fixes issue #187

### DIFF
--- a/README.md
+++ b/README.md
@@ -647,6 +647,22 @@ Also supports [additional options](http://www.elasticsearch.org/guide/en/elastic
 City.search "san", boost_by_distance: {field: :location, origin: [37, -122], function: :linear, scale: "30mi", decay: 0.5}
 ```
 
+### Routing
+
+Searchkick supports elasticsearch's routing feature.
+
+```ruby
+class Contact < ActiveRecord::Base
+  searchkick routing: :user_id
+end
+```
+
+Reindex and search with:
+
+```ruby
+Contact.search "John", routing: current_user.id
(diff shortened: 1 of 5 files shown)


bogdanovich:
@ankane 
Is there any issues that prevent this commit to be merged?

ankane:
Thanks for the pull request Anton! Sorry for the **really** slow response. This is a great start.  A few things that would be helpful:

- an option to make the mapping easier
- how to use it in the readme
- tests if possible

bogdanovich:
@ankane Done.
Getting some weird NoBrainer failures on Travis though.

bogdanovich:
Tests are passing with nobrainer 0.19.0 and failing with nobrainer 0.23.0,
so yeah this is not related to routing feature..

ankane:
Thanks @bogdanovich, merged!

PR #380, opened 2015-02-02
Add ability to reverse where using not filter

anfinil:
Adds ability to create negative filters: for example `{not: {location: {geo_distance:....` , `{not: {user_id: {in: [1,2]}}}`

It is very useful to reverse where queries.


ankane:
Hey @anfinil, thanks for the PR 👍  Sorry we couldn't get this in sooner. The latest version of Searchkick now supports a `_not` option to do this.

Issue #356, opened 2015-01-05
Retain x number of indexes on swap.


incorvia:
At my company, we ran into an issue that caused a partially created index to be swapped into production. Since old indexes are deleted on swap, we had to do a full-reindex to recover which took half an hour. Our thinking going forward is we'd like to perform the reindex and save the last non-aliased index and delete all others.  

Obviously SearchKick does not support this. It seems that in order to support it some modifications would be needed to be made to the logic near here:

https://github.com/ankane/searchkick/blob/master/lib/searchkick/index.rb#L139

Probably would need to sort the non-aliased indexes by date before deleting old indexes and delete all but x number which could be set in a setting like:

  searchkick retain: 1

Would such a feature / approach be welcomed?

incorvia:
@ankane, thanks for addressing my issues.. While I don't think that (retain: true) is really what I was hoping, I'd say retain: :previous is.. what I was hoping for.. I'm not sure why this would be confusing:

Product.reindex(retain: previous) # indexes: [A] (Initial, nothing to retain)
Product.reindex(retain: previous) # indexes: [A,B] (swaps, and retains A)
Product.reindex(retain: previous) # indexes: [B,C] (swaps, and retains B)

For me I'd just be ok with retain_previous: true.. but perhaps you'd prefer retain: true, false, or :previous? thoughts?

ankane:
In some circumstances, people may not want to clean indices at all, so I was thinking that option should be included in this feature as well.

incorvia:
@ankane, I'm certainly ok with that.. before we proceed.. Would you be ok exposing this option to the rake task as well via an ENV?

ankane:
See #381

PR #381, opened 2015-02-02
Add option for retaining indexes to Model#reindex

incorvia:
I want to resolve issue #356 regarding adding an option to retain indexes to Model#reindex.  In it's current form this PR is not complete as it does not feature tests, nor the ability to utilize this feature in searchkick index tasks.  I just wanted to get feedback on the strategy.

@ankane reindex_scope seemed overly complicated and I simplified it a little bit.. I may have missed something but this seems to work as expected.. wondering if I can get your thoughts.

### DIFF
--- a/lib/searchkick/index.rb
+++ b/lib/searchkick/index.rb
@@ -137,42 +137,33 @@ module Searchkick
     end
 
     # remove old indices that start w/ index_name
-    def clean_indices
+    def clean_indices(retain = false)
       all_indices = client.indices.get_aliases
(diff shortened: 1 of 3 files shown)


incorvia:
@ankane I did add a test here.

ankane:
Hey @incorvia, I went ahead and added `retain: true/false` option. I'm still worried other options (even `:previous`) will be confusing when you consider failed reindexes, so think we should close this for now. At the very least, with `retain: true`, you can now add this functionality to your app.

```ruby
# in model or concern

def reindex
  custom_clean_indices
  searchkick_reindex(retain: true)
  custom_clean_indices
end

def custom_clean_indices
  # use searchkick_index.all_indices to do what you want
end
```

Thanks for taking the time to discuss / fork and making Searchkick better  🎉

PR #406, opened 2015-03-07
fix typo in readme (was "body" instead of "query")

adrienbourgeois:


ankane:
Thanks @adrienbourgeois - however, `body` now is preferred over `query`.

PR #418, opened 2015-03-30
Update badges

abookyun:


ankane:
Thanks for the pull request David!  However, I don't think more badges are necessary.

abookyun:
Perhaps the svg commit only is fine?

ankane:
Sounds good, but let's keep it in the same place.

abookyun:
Sure, right away.

ankane:
Thanks David!

Issue #421, opened 2015-04-09
calling 'map' method on Searchkick::Results object returns always same order


bgvo:
I have a User model with following spec:

```
def search_data
    {
        full_name: full_name,
        published: published,
        sn: sn
    } 
end
```

Then I try to get ordered results on `sn` field. This is my query:

`User.search("pablo", order: {sn: :asc}, fields: [:full_name], limit: 10)`

Inspecting the resulting object reveals that it is working as expected:

`User.search("pablo", order: {sn: :asc}, fields: [:full_name], limit: 10)` gives:

```
#<Searchkick::Results:0x007feae41a3958
 @klass=User,
 @options={:page=>1, :per_page=>10, :padding=>0, :load=>true, :includes=>nil, :json=>false},
 @response=
  {"took"=>75,
   "timed_out"=>false,
   "_shards"=>{"total"=>1, "successful"=>1, "failed"=>0},
   "hits"=>
    {"total"=>5,
     "max_score"=>nil,
     "hits"=>
      [{"_index"=>"users_production_20150409165753551", "_type"=>"user", "_id"=>"551598ba32663300160d0000", "_score"=>nil, "sort"=>[1]},
       {"_index"=>"users_production_20150409165753551", "_type"=>"user", "_id"=>"54e3ba1d6538630003090000", "_score"=>nil, "sort"=>[2]}]}}>
```

You can see that first item in `hits` is the one supposed to be. If I run the query specifying descending order then first `hits` item is the one with `sn` 2. However, when I call `#map` method on Searchkick::Results object it always returns the same order not matter the order specified.

Is anything I'm missing?

Thanks!

ankane:
Hey Borja, the map method should retain the order in `hits`.  Can you create a failing test case for it?

bgvo:
Hi! I made the test (see above) and everything is fine. Please, leave the issue open until I see what happens.

Thanks!

bgvo:
My BIG (did I say BIG?) fault. I was mismatching names (I guess), because it is working perfectly.

However, there you have another test to add ;)

Thanks

PR #422, opened 2015-04-13
Test it's ok - #421

bgvo:
Regarding issue #421. I created the test and it's passing.


ankane:
Thanks @bgvo. However, I'm not sure this test is needed.

PR #436, opened 2015-04-30
Add an option to enable transpositions in fuzzy search

parkdy:
This change allows transpositions (switching two characters) to be counted as a single operation when calculating the edit distance. For example:

User.search(''Jonh", { misspellings: { edit_distance: 1, transpositions: true } })

would return "John" as a match.


ankane:
Hi David, thanks for the pull request!  Bad news: I didn't see this one until now (sorry!).  Good news: this feature is now available on the master branch.

parkdy:
Thanks ankane, glad to see this change make it into Searchkick!

PR #438, opened 2015-05-03
Add new Ruby 2.2 version

nicolasleger:


ankane:
Thanks for the pull request Nicolas!  In the interest of saving resources for Travis CI, I don't think we need to test on both 2.1 and 2.2 - they should be the same for everything Searchkick is doing (nothing fancy).

nicolasleger:
There could be minor changes between versions (see https://www.ruby-lang.org/en/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/).
But if you want to keep only one version tested, it would be better to use the newest stable one `2.2`.

ankane:
Bumped travis to use 2.2.

PR #439, opened 2015-05-06
Add range option to boost filters.

ezekielriva:
I've trying to boost results beetween two values without
    success, so I added two simple lines to allow that feature.
    Now, you could boost results beetween two values sending
    a Range object as parameter.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -547,6 +547,8 @@ module Searchkick
         else
           {in: {field => value}}
         end
+      elsif value.is_a?(Range)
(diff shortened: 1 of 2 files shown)


ezekielriva:
Hi @ankane I've been ran the test several times and they didn't fail. Is something wrong in Travis?

ankane:
Thanks for the pull request @ezekielriva! :+1: 

The tests were failing because the `2..9` range boosts both Tomato B and Tomato C, so sometimes it would return the latter.  I've expanded this to allow for any where condition to be used with boost_where.

PR #449, opened 2015-05-26
Make `Searchkick.disable_callbacks` threadsafe

stormsilver:
Switch from a class variable to a fiber-local variable. Note that this only applies to disabling/enabling global callbacks, not per-model callbacks (those are still not threadsafe).

### DIFF
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -24,13 +24,11 @@ module Searchkick
   class InvalidQueryError < Elasticsearch::Transport::Transport::Errors::BadRequest; end
 
   class << self
-    attr_accessor :callbacks
     attr_accessor :search_method_name
     attr_accessor :wordnet_path
     attr_accessor :timeout
     attr_accessor :models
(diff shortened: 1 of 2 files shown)


lib/searchkick.rb
```diff
@@ -24,13 +24,11 @@ class UnsupportedVersionError < StandardError; end
   class InvalidQueryError < Elasticsearch::Transport::Transport::Errors::BadRequest; end
 
   class << self
-    attr_accessor :callbacks
```
stormsilver: Not sure about getting rid of this - nothing in the gem appears to use it, but I don't know if people using the gem would be checking its value. You could argue that this should always have been a private implementation and the public API is to use `callbacks?`, and therefore anyone inspecting `callbacks` is _wrong_. You could also just `alias :callbacks :callbacks?`

ankane:
Great, thanks Eric!  I think getting rid of `attr_accessor` is fine.

PR #454, opened 2015-06-02
[Facets] Support overriding field in facet search

meetrajesh:
Currently Searchkick defaults to the facet hash key to determine the field it should aggregate on. This commit allows someone using facets search to specify the exact elastic field they want the facet to aggregate on. This is useful if the name of the facet key doesn't match the field they want to search on.

It also benefits people that want two separate aggregations (i.e. with different where clauses) on the same field. Currently this is not possible since the hash keys would have to be the same, therefore causing one to override the other. With this commit, the 2nd hash key can be something different but the field aggregated is still the same.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -301,7 +301,7 @@ module Searchkick
             else
               payload[:facets][field] = {
                 terms: {
-                  field: field,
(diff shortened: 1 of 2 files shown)


ankane:
Awesome, thank you sir.

PR #455, opened 2015-06-04
Aggregations support

igorbelo:
since facets are deprecated, and will be substituted for aggregations, i've implemented for the search method, to accept aggregations options, which is very similar to facets.

it would be bad if we lost the searchkick friendly query, making a custom request to get aggregated data (with aggregations instead of facets).

I'll write some docs after considerations

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -368,6 +368,49 @@ module Searchkick
           end
         end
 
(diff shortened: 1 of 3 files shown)


igorbelo:
As aggregations were added to Searchkick 1.0, this PR is not necessary anymore.

Issue #458, opened 2015-06-10
Range query doesn't work for :id field 


bogdanovich:
Searchkick converts id to _id which is incorrect, because id field could be a long value and _id is always a string.

``` ruby
def where_filters(where)
  filters = []
  (where || {}).each do |field, value|
  field = :_id if field.to_s == "id"  # why??
```

``` ruby
# returns nothing
Model.search("test", where: { id: { gt: 123455 }})
```

Current code generates the following query  that returns nothing 
`{"range":{"_id":{"from":1099511627779,"include_lower":false}}}`

It should generate the following query to return proper results
`{"range":{"id":{"from":1099511627779,"include_lower":false}}}`

ankane:
Going to close. No plans to change how `id` is handled right now.

PR #460, opened 2015-06-11
Fix range queries for "id" field

bogdanovich:
Fixes issue #458


bogdanovich:
Breaking changes:
If id field is not indexed, :_id should be used to get record by id
```ruby
Product.search "*", where: {_id: 12342}
```

bogdanovich:
There is an alternative PR #461 which doesn't have breaking changes

ankane:
Cleaning up PRs

PR #461, opened 2015-06-11
Fix range queries for "id" field  (alternative)

bogdanovich:
Fixes issue #458

This is alternative to PR #460 
It doesn't introduce breaking changes

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -467,7 +467,6 @@ module Searchkick
     def where_filters(where)
       filters = []
       (where || {}).each do |field, value|
-        field = :_id if field.to_s == "id"
 
         if field == :or
           value.each do |or_clause|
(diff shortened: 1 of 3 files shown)


ankane:
Hey @bogdanovich, thanks for the PR 👍  The `_id` column is indexed as a string, so sorting will be off (`"11" < "2"`). I don't think there's a way around this besides using a different column name than `id`, since Searchkick treats it specially.

MaxDBN:
Hey, I can't search using a condition like 
where:{_id:{lt:100}} or where:{id:{lt:100}}.
Can you please tell me if there is something I could do wrong? @ankane

ankane:
@MaxDBN You'll need to use a field with a name other than `id` for range queries. In Searchkick 2.0, it probably makes sense to throw an error if you try to do this with an explanation.

MaxDBN:
@ankane Bummer. I was hoping I could use it to search my db in batches of 10K results. I guess no way of doing that with Searchkick without increasing max_result_window?

bogdanovich:
@MaxDMN, we had to patch searchkick and use forked version for returning paginated results in a way 
'Where id > last_id limit 1000".  Basically what required is decoupling  _id and id and treating them separate

PR #465, opened 2015-06-23
Additional support for regular expressions (using strings instead of Regexp object)

meetrajesh:
Currently Regexp Ruby objects aren't serializable using the msgpack gem that we (Instacart) use to serialize objects to pass into InstacartRPC calls. Ideally we'd pass in the query json as a string across the wire, but right now, we're passing the params directly as is.

Currently, the only way to perform a regexp filter query against ES using Searchkick is to use a Regexp Ruby object. This PR changes that and allows someone to perform regexp searches using plain strings.

Where in the past you would have to say:

    {where: {field: /pattern/}}

Now you can say:

    {where: {field: {regexp: "pattern"}}}

Let me know what you think.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -504,6 +504,8 @@ module Searchkick
                     }
                   }
                 }
+              when :regexp # support for regexp queries without using a regexp ruby object
+                filters << {regexp: {field => {value: op_value}}}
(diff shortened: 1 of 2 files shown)


moelee:
+1

ankane:
Cool, thanks!

PR #467, opened 2015-06-29
Create write and read aliases during reindex_scope

naiyt:
This PR changes the functionality of `reindex_scope`. With these changes during a reindex you will have your `write` alias pointing to both the new and old index. That way if you are adding new entries during a long reindex you can write it to both the old and new index. This should help improve zero downtime reindexing, because someone can still search for that new entry inside of the old index while you are reindexing. (Otherwise if someone adds an entry they won't be able to search it until your reindex completes.)

You can utilize this by iterating through the indices that the write alias points to when saving a new entry.

-  `reindex_scope` gets called during reindexing
-  A new index gets created with the current timestamp in the name
- We set the write alias to also point to the new index by calling `reindex_scope`
- If we do not have any old indices (this is our first index) we point the read alias to the new index as well
- The reindex happens
- We remove the write alias from the old index and move the read alias to the new index

### DIFF
--- a/lib/searchkick/index.rb
+++ b/lib/searchkick/index.rb
@@ -23,19 +23,8 @@ module Searchkick
       client.indices.refresh index: name
     end
 
-    def alias_exists?
-      client.indices.exists_alias name: name
-    end
-
-    def swap(new_name)
-      old_indices =
-        begin
-          client.indices.get_alias(name: name).keys
-        rescue Elasticsearch::Transport::Transport::Errors::NotFound
-          []
-        end
-      actions = old_indices.map { |old_name| {remove: {index: old_name, alias: name}} } + [{add: {index: new_name, alias: name}}]
-      client.indices.update_aliases body: {actions: actions}
+    def alias_exists?(alias_name=name)
+      client.indices.exists_alias name: alias_name
     end
 
     # record based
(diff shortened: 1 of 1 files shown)


ankane:
Hi Nate, thanks for the pull request! Unfortunately, I don't think it's possible to write to multiple indices at the moment. https://github.com/elastic/elasticsearch/issues/6240

naiyt:
Hey Andrew, thanks for the reply and for maintaining searchkick.

The strategy this allows isn't actually to write to the alias (and thus both of the indices) like that elasticsearch issue mentioned. It's to give you an easy reference to both your old and new indices during reindexing so that you can write to them both manually in your app.

Pseudocode for it might be:

```ruby
def index_user(user_to_index)
  # find_alias is a helper method to find a specific alias in the format
  # {"alias_name" => [old_index, new_index]}
  write_indices = find_alias("users_write")
  write_indices.each do |index|
      index.import user_to_index
  end
end
```

That way when a new object gets created you won't get stale results during reindexing. Our reindexing can take quite a long time and without something like this any new objects created during reindexing won't be searchable until it's finished. We've been using this strategy with tire for awhile.

We may not actually end up moving to searchkick right now for a few reasons. If you think this might be helpful for others let me know, and I'll see if I can fix the broken specs.

naiyt:
We've gone with a different solution, so I'm going to close this.

talharamati:
Hi @naiyt !
Can you please collaborate what was your final solution?

Thanks :)

Tal.

naiyt:
Hey @talharamati, we ended up sticking with Tire for now, because there were various issues related to our implementation that weren't working will with Searchkick. If you're looking to do this exact thing, you should be able to use the changes in this PR, though.

talharamati:
Ok thanks, I'll try it out.

My problem is that while I reindex my data in production, any write action goes to the old index, so the new one isn't up do date (In the meanwhile people updating, deleting and creating documents).
I want to be able to write to both indices and read from the old one while reindexing.

If someone has any suggestion, it would be great

narendraj9:
@talharamati I have a similar problem. What did you choose to do about this?

PR #472, opened 2015-07-09
Add support for fuzzy transpositions

meetrajesh:
Allows users to misspell words by transposing two letters and having it still work. Common example is lasagna vs. lasanga. This is normally not caught when the edit distance is set to just 1.

(cc: @richardking)

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,13 @@ You can also change the edit distance with:
 Product.search "zucini", misspellings: {edit_distance: 2} # zucchini
 ```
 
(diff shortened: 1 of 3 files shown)


ankane:
I'm afraid this will break backwards compatibility.  A safer approach would be to make it an option:

```ruby
misspellings: {transpositions: true}
```

When Searchkick hits 1.0, we can make this the default.

meetrajesh:
@ankane Good point. How does it look now? I've made it an option like you suggested.

ankane:
:+1: Looks great - big win for misspellings.

PR #473, opened 2015-07-10
add multiplier boost

richardking:
thoughts on this @ankane? cc @meetrajesh 

basically ops wants to experiment with lowering scores of results, based on various factors. the current `boost_by` doesn't quite work because it adds up all the function scores, and then multiplies the score by that sum.

i can add to readme, and anything else you suggest, but wanted to run it by you guys first.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -238,6 +238,40 @@ module Searchkick
           }
         end
 
+        multiply_filters = []
+
+        multiply_by = options[:multiply_by] || {}
+        if multiply_by.is_a?(Array)
+          multiply_by = Hash[multiply_by.map { |f| [f, {factor: 1}] }]
+        end
+
+        multiply_by.each do |field, value|
+          script_score =
+            if below12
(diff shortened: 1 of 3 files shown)


ankane:
This is a tough one.  I'm worried it'll be confused with `boost_by`.  One possible option would be to add:

```ruby
boost_mode: "multiply", boost_by: {found_rate: {log: false}} # log = use natural log
```

Thoughts?  This is another change that may make sense as a Searchkick 1.0 default.

richardking:
Agreed it's a little confusing with boost_by. I'll play with it a bit to see if I can just modify boost_by as you suggested

richardking:
this commit is still kind of rough but i wanted to get the gist of it out. it seems like if we add `boost_mode` alongside boost_by, then we could only do either multiply or sum, does that sound right? we would still want to keep the boost_by sums, but also add multiply. i think this should keep backwards compatibility but i still have to play with edge cases. lmk what you think.

ankane:
Looks good.  For reference, the formula is:

```ruby
(_score + boost_mode: sum) * boost_mode: multiply
```

Issue #480, opened 2015-07-23
Add search results records method (to provide a scope instead of an array)


fabn:
It would be very useful to expose a `#records` method in returned results. You [already building a scope](https://github.com/ankane/searchkick/blob/master/lib/searchkick/results.rb#L147) for returned records. However you call `#to_a` on it and this trigger record loading.

Have a scope instead of an actual array has lot of benefits, some of them:
- Lazy loading, query won't be executed until you need results. This can be very useful to deal with cache keys, e.g. `<%= cache results.records.maximum(:updated_at) do %>`
- Results refinements (as you pointed in [this comment](https://github.com/ankane/searchkick/issues/140#issuecomment-36440598))
- Selective loading of retrieved fields. Suppose you have some big texts in database for returned results, one can use the awesome `pluck` method to not load all the other stuff. See the following example.

``` ruby
def autocomplete
  # This will only retrieve id and name attributes instead of whole records.
  results = Whatever.search(params[:query], fields: [{name: :text_start}], limit: 10)
  render json: results.records.pluck(:id, :name) # or even better results.pluck(:id, :name) if possible
end
```

Mavvie:
The issue with implementing this is single table inheritance, specifically here https://github.com/ankane/searchkick/blob/master/lib/searchkick/results.rb#L37-L39

Say you have a `class Animal` with subclasses `Dog`, `Cat`, `Lizard`.

There's two solution on my mind:

1. Keep it as a relation of the parent class (`Animal`). This kinda sucks because you have to separate them yourself
2. Return a hash of the different types (`record['animal']`, `records['dog']`, etc). This kinda sucks if you want all of them together.

I'm leaning toward `1.`, any thoughts?

Here's an example of `2`: #489.

fabn:
> Keep it as a relation of the parent class (Animal). This kinda sucks because you have to separate them yourself

I seldom used STI in rails, majority of times concerns are a better choice. 

In any case given you have a scope for `Animal` model imho the best choice is to return as is (i.e. `Animal.where(id: [list, of, returned, ids])`) because then the user can narrow the scope to suit its needs and once the scope is resolved (by doing the actual query) objects will be returned as subtypes, e.g.

```ruby
Animal.search('...').records.to_a.map(&:class) # => [Dog, Cat, Lizard]
```

Also if one uses STI usually there is some shared behaviors that is reusable across subclasses thus if one just needs animal names one can do `Animal.search(...).pluck(&:name).

[...]

PR #490, opened 2015-08-04
Implement results#records method for ActiveRecord

Mavvie:
#480


ankane:
Awesome, thanks Tyler!

PR #292, opened 2014-09-28
Add options for query : min_score, facets/all_terms, facets/field

sebfie:
Hello,

I added options to query : 
- min_score => Required when we want to filter pertinence results 
- facets/all_terms => Allow you to pass the all_terms (bool) param for facets
- facets/field => You can now specify the field for facet. 

I did this fix because I migrated my project to searchkick and needed this options. I am not an expert of ES, so, be critical, I maybe made huge mistakes !

PR #492, opened 2015-08-07
Add min_score option, with result test!

st0012:
I found the feature in #292 would be very helpful to me, so I add result test and rebase it. Hope this PR can be merge ASAP, thanks 😁.

### DIFF
--- a/test/query_test.rb
+++ b/test/query_test.rb
@@ -10,4 +10,16 @@ class QueryTest < Minitest::Test
     assert_equal ["Apple", "Milk"], query.map(&:name).sort
     assert_equal ["Apple", "Milk"], query.execute.map(&:name).sort
   end
+
+  def test_with_effective_min_score
+    store_names ["Milk", "Milk2"]
+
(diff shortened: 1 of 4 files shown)


st0012:
@ankane Hello ~ is there any problem in my code?

ankane:
Same ask as #292 - can you add tests on how these options affect the search results rather than how they modify the request?

test/query_test.rb
```diff
@@ -11,4 +11,15 @@ def test_basic
     assert_equal ["Apple", "Milk"], query.execute.map(&:name).sort
   end
 
+  def test_with_effective_min_score
+    store_names ["Milk", "Milk2"]
+
+    assert_equal ["Milk"], Product.search("Milk", min_score: 0.1).map(&:name)
```
dlackty: @ankane I thought this test and the following one shows "how these options affect the search results"? cc @st0012

ankane: These tests do, but the facet and sql tests do not.

st0012: @ankane I added facets' result test, but I am not sure what you want to test in sql test. Won't it be a lot like query test?

PR #495, opened 2015-08-15
add bulk update method

richardking:
we have a use case where we want to update a single field for a lot of records at a high frequency, and this should increase the number of documents we can send to the bulk endpoint and speed up the indexing process.

### DIFF
--- a/lib/searchkick/index.rb
+++ b/lib/searchkick/index.rb
@@ -70,6 +70,16 @@ module Searchkick
       end
     end
 
+    def update(records, fields)
+      records.group_by { |r| document_type(r) }.each do |type, batch|
+        client.bulk(
+          index: name,
+          type: type,
+          body: batch.map { |r| {update: {_id: search_id(r), data: {doc: search_data(r).slice(*fields)}}} }
+        )
+      end
+    end
+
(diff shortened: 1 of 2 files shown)


ankane:
@richardking thanks! have you had a chance to benchmark this? i'm not sure it'll speed up reindexing by much due to the way [partial updates work](https://www.elastic.co/guide/en/elasticsearch/guide/current/partial-updates.html) and the fact that Searchkick needs to construct all `search_data` fields.

richardking:
hmm, thanks Andrew. i was aware about the partial updates but was thinking that it would at least cut down on the amount of data we need to send over the wire, since there's a lot of data for each doc. but it does seem like the biggest win would be not having to create the `search_data` fields. will see what we can do about that.

ankane:
Hey @meetrajesh, here's the PR I mentioned earlier. To keep from generating `search_data`, Searchkick could provide a `partial_reindex` method that could be used like:

```ruby
Product.partial_reindex(:search_pricing)
```

with:

```ruby
class Product
  def search_pricing
    {
      price: price
    }
  end
end
```

ankane:
Added proposal in #766

PR #504, opened 2015-08-25
Disable cutoff_frequency when misspellings is false

### LINKED ISSUE #339: Operator "or" ignored when misspellings set to false.
We're having an issue with the "or" operator when setting misspellings to false:

This works as expected.
```ruby
Link.search (params[:query] || "*"),
  operator: "or",
  where: {
    published_at: {lte: Time.zone.now}
  },
  page: params[:page], per_page: (params[:per_page] || 20),
}
```
However, as soon as we add `misspellings: false`, `operator: "or"` seems to be ignored causing the operator to default back to "and".

toyray:
When cutoff_frequency is specified with misspellings = false, the high frequency terms are no longer matched against (sort of becomes an "and" query even when operator "or" is specified). 

This is most likely to happen mainly when searching against fields that share similar values e.g a field containing tags associated with the current object.
 
Added a couple of tests for this and opted for disabling cutoff_frequency instead of providing option for user to specify a custom cutoff_frequency.

This might fix #339 as well, not 100% sure without information from the original poster.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -96,11 +96,6 @@ module Searchkick
               }
 
               if field == "_all" || field.end_with?(".analyzed")
-                shared_options[:cutoff_frequency] = 0.001 unless operator == "and"
-                qs.concat [
-                  shared_options.merge(boost: 10 * factor, analyzer: "searchkick_search"),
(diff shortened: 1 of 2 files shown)


ankane:
Thanks Roy, great find, fix, and tests!

PR #505, opened 2015-08-25
Add prefix_length option for misspellings

### LINKED ISSUE #410: fuzziness on short words
fuzziness should have a prefix length of 2 or, better, be customizable; it replaces things like "api" -> "any" which really isn't great.

http://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-fuzzy-query.html

toyray:
Added prefix_length option for misspellings to disable fuzzy matching for short words (keeping prefix_length for lack of a better name).

Main gotcha is when a query term has the same length as prefix_length whereby it becomes an exact match query, hence the extra test case.

This should fix #410, but defaulting prefix_length to 0 if unspecified for backwards compatibility.

Let me know if you want me to submit  PR to update README for this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,18 @@ Product.search "mikl", misspellings: {transpositions: true} # milk
 
 This is planned to be the default in Searchkick 1.0.
 
+For short queries, it may be helpful to turn on mispellings after a specified number of letters.
+
+```ruby
+Product.search "api", misspellings: {prefix_length:2} # api, apt, not any
+```
+
+**Note:** If query length is the same as `prefix_length`, misspellings is automatically turned off.
+
+```ruby
+Product.search "ah", misspellings: {prefix_length:2} # ah, not aha
+```
+
 ### Indexing
 
 Control what data is indexed with the `search_data` method. Call `Product.reindex` after changing this method.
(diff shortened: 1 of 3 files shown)


ankane:
Thanks Roy, this is useful!  The case when prefix_length is the same length as the term seems counter-intuitive (seems like `ap` with prefix_length of 2 should match `api`), but I'm not sure there's an easy way around it.  An example in the readme would great, and it'd be good to mention this caveat.

toyray:
I'll see if there's a way to tweak the prefix_length based on the search term to make the behavior more intuitive.

Anyways, I'll add a commit to document this in the README and rebase this for easier merging later.

toyray:
I added a way to handle the case where prefix length = term length by using the term length - 1 so the misspellings still work for those cases. This should also work for the "or" operator since the length of the shortest token is always used.

Updated the test cases accordingly (let me know if there are other cases that should be covered)

I'm not too sure if below is the best way of getting the shortest token from the query term.

```ruby
search2_prefix_length = [
    (searchkick_index.tokens(term, analyzer: "searchkick_search2")
        .min_by(&:length)
        .try(:length) || 1
    ) - 1,
     prefix_length
].min
```

the other alternative I had was

``` ruby
search2_prefix_length = [
    term.split(" ").min_by(&:length).try(:length) || 1) - 1,
     prefix_length
].min
```

Let me know what you think and if this is good, I'll add another commit to update the README

ankane:
With the `prefix_length - 1` approach, `ak` will match `ap` with a prefix length of 2, which is also counterintuitive.  I'd say we go with the original approach.

toyray:
Yup, I'll rebase without the `prefix_length - 1` commit and add a commit to update the README.

Thanks for the feedback, Andrew.

ankane:
Merged, thanks @toyray!

PR #506, opened 2015-09-01
Searchkick env

muffy:
In order to use staging and not have to run the local catalog server, it's necessary to tell search that it's pointed at and running on staging. We can set the ELASTICSEARCH_URL to be the staging url, but then we need to also use "staging" as the environment.

### DIFF
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -63,7 +63,7 @@ module Searchkick
   end
 
   def self.env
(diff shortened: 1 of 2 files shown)


ankane:
What if instead Searchkick provided a setter for the env? Then you could do:

```ruby
Searchkick.env = ENV["ELASTICSEARCH_ENV"] if ENV["ELASTICSEARCH_ENV"]
```

This would be more flexible for Searchkick tests.

muffy:
Works for me.

ankane:
Cool, done!

PR #509, opened 2015-09-07
partial and bulk updates

talharamati:
We use these for updating a part of one or more documents without generating the entire search_data.

@ankane, let me know if that makes sense and I'll add specs and docs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
   - elasticsearch
   - mongodb
 before_install:
-  - ./ci/before_install.sh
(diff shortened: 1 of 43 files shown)


richmolj:
@talharamati what's the status of this? Partial updates would be super helpful to me.

ankane:
Hey @talharamati, this seems useful. Something seems off with the diff (43 files changed). Can you update?

ankane:
Cleaning up PRs. Added a proposal for partial updates in #766

PR #513, opened 2015-09-11
Add better error message when using smart_facets on facets

klebervirgilio:
Hi,

I looks like if you use `:facets + smart_facets => true`, searchkick gem will raise an exception.
```ruby
NoMethodError: undefined method `reject' for nil:NilClass
from: .../searchkick-0.9.0/lib/searchkick/query.rb:315
```

Was wondering, is it ok to raise an explicit Exception? or.. Would u rather set a default empty `Hash`?

Cheers,
Kleber

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -311,7 +311,7 @@ module Searchkick
             # offset is not possible
             # http://elasticsearch-users.115913.n3.nabble.com/Is-pagination-possible-in-termsStatsFacet-td3422943.html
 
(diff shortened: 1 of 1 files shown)


ankane:
Thanks @klebervirgilio, I like the empty hash strategy.

klebervirgilio:
Thanks @ankane, for the feedback! I fixed and squashed. Does it look better now?

ankane:
Modifying options could have implications downstream, so it'd be safest to do

```ruby
deep_merge!(where: options[:where] || {}).reject ...
```

klebervirgilio:
@ankane, how about now?

ankane:
:+1: thanks!

PR #522, opened 2015-09-29
Strip status code from ES error message

mehulkar:
This leaves only the JSON response and makes it easier to parse e.message
higher up the stack.

After this, my code can look roughly like this:

```ruby
begin
  render json: Product.search('*')
rescue InvalidQueryError => e
  render json: JSON.parse(e.message)
end
```


ankane:
Hey @mehulkar, thanks for the PR 👍  I think this should be left to the application to do.

PR #527, opened 2015-10-07
Add nested object filter support

lucasbibiano:
Add basic support for nested object filter.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -264,6 +264,17 @@ module Searchkick
 
         # filters
         filters = where_filters(options[:where])
+
+        # nested filter
+        if nested_filter_value = options[:nested]
+         filters << {
+           nested: {
+             path: nested_filter_value[:path],
+             filter: { and: where_filters(nested_filter_value[:where]) }
+           }
+         }
+        end
+
         if filters.any?
           if options[:facets]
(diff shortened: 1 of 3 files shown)


ankane:
Hi @lucasbibiano, thanks for the pull request!

What if we added first class support for nested objects?

```ruby
searchkick nested: [:creator]
```

and

```ruby
Product.search term, fields: ["creator.first_name"]
Product.search term, where: {"creator.first_name" => "John"}
```

ankane:
Hey @lucasbibiano, any thoughts on adding first class support?

ankane:
Cleaning up PRs

eddietejeda:
The PR is fairly old and can't be merged as it is, but I managed to merge some of the code here for testing purposes here: https://github.com/eddietejeda/searchkick/commits/master

And it seems to be working well. I have been using it in a large application (in a development environment) for a couple weeks and I don't see any issues.

@ankane Is there a reason why this is not merged into the main branch? Are there any concerns you have with supporting this now? Or are you just pacing how features are added?

If it's just a question of having someone work on it, I would be glad to clean up @lucasbibiano's code a bit.  Thoughts?

Re: first class support. While that would be nice, I find the way it's implemented in this PR works well for our use case.

PR #531, opened 2015-10-13
Allow to pass synonyms as Proc

sharshenov:


ankane:
Thanks @sharshenov!  I think it'd be best to check if `synonyms.respond_to?(:call)`.  Also, can you leave the existing line unmodified?

sharshenov:
Fixed

ankane:
Awesome, thanks! :+1:

PR #536, opened 2015-10-22
Add support for aggregations

meetrajesh:
This PR adds support for aggregations to searchkick as an alternative to facets. Moreover, since facets will be [removed](https://www.elastic.co/guide/en/elasticsearch/reference/2.0/_removed_features.html) from Elasticsearch as of v2.0, people who were previously using facets in searchkick will need to move to aggregations if they want to use it with ES v2.0.

This PR attempts to keep the aggregations behavior as similar as possible to facets so people will not have to change too much code when migrating their searchkick clients.

I went back and forth between "aggs" and "aggregations" in the code (elasticsearch supports both), and eventually opted for "aggs" everywhere for standardization and also because "aggregations" is more typo-prone.

The searchkick query looks like this:

`Product.search("Product", aggs: {store_id: {where: {in_stock: true}}})`

which generates an ES query payload that looks like this:

```
{
   "query": {
   },
   "size":10,
   "from":0,
   "aggs": {
      "store_id": {
         "filter": {
            "bool": {
               "must": [
                  {
                     "term": {
                        "in_stock": true
                     }
                  }
               ]
            }
         },
         "aggs": {
            "store_id": {
               "terms": {
                  "field": "store_id"
               }
            }
         }
      }
   }
}
```

which produces an ES response that looks like this:

```
{
   "store_id": {
      "doc_count": 1,
      "store_id": {
         "doc_count_error_upper_bound": 0,
         "sum_other_doc_count": 0,
         "buckets": [
            {
               "key": 1,
               "doc_count": 1
            }
         ]
      }
   }
}
```

Calling `response.aggs` (i.e. `Product.search(...).aggs`) returns a hash that looks like:

```
{
   "store_id": {
      "doc_count": 1,
      "doc_count_error_upper_bound": 0,
      "sum_other_doc_count": 0,
      "buckets": [
         {
            "key": 1,
            "doc_count": 1
         }
      ]
   }
}
```

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -266,7 +266,7 @@ module Searchkick
         # filters
         filters = where_filters(options[:where])
         if filters.any?
-          if options[:facets]
+          if options[:facets] || options[:aggs]
             payload[:filter] = {
               and: filters
             }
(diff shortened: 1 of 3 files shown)


ankane:
Thanks @meetrajesh, this feature would be great. I'll take a look when I have a chance.

meetrajesh:
Hi Andrew, have you had a chance to take a look?

ankane:
Great, thanks Raj! :+1: on aggs over aggregations (less typing!) and the format of results.  Will release and add to the readme - officially deprecating facets - once support for Elasticsearch 2.0 is in place (ideally tonight, but looks to be more work than anticipated)

matcouto:
G'day Andrew, really good improvements since I last stopped by here. Good on ya! :smiley: 
I was just wondering whether you'll be releasing **1.0.1v** anytime soon?
Cheers

ankane:
1.0.1 just released.

PR #547, opened 2015-11-06
Callable language param in searchkick

sld:
If we have index for each language it seems necessary to have stemmer for the lang.

This pull request adds support of lambda `language` param in `searchkick`.

Now we can write something like this:
```
searchkick index_name: -> { searchkick_index_name(I18n.locale) },
           language: -> { searchkick_language(I18n.locale) }
```

### DIFF
--- a/lib/searchkick/index.rb
+++ b/lib/searchkick/index.rb
@@ -201,6 +201,8 @@ module Searchkick
 
     def index_options
       options = @options
+      language = options[:language]
+      language = language.call if language.respond_to?(:call)
 
       if options[:mappings] && !options[:merge_mappings]
         settings = options[:settings] || {}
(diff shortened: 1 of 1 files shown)


ankane:
Looks good.  Thanks @sld!

fabn:
Multi language is always an issue. How does this works precisely? If i understand the code `index_options` is only called when a full reindex is issued, and index is recreated from scratch.

If so what's the use case of this PR? According to the example above the only use case I see is to recreate index with a callable (e.g. `I18n.with_locale(:it) { SomeModel.reindex }`) but I don't get the point.

If I have a multi language model, e.g.

```ruby
class Article < ActiveRecord::Base {
                     :id => :integer,
                  :title => :text,
                   :body => :text,
                 :locale => :string, # one I18n.available_locales
}
```

this patch is not useful, because cannot be used as in 

```ruby
article = Article.find(1)
I18n.with_locale(article.locale) { article.reindex } 
```

Am I right?

ankane:
If you use a proc for the index name and proc for the language (as in the example above), you can create separate indices for each language.

PR #554, opened 2015-11-15
add reminder to start elasticsearch

steve-chen-nyc:
It's easy for a beginner to install elastic search but forget to start it. This would be a nice reminder.


ankane:
Thanks for the suggestion Steve!

PR #560, opened 2015-11-23
Mongoid bug

benjaminharel:
`records.respond_to?(:primary_key) && records.primary_key` 
returns true for mongoid.

### DIFF
--- a/lib/searchkick/results.rb
+++ b/lib/searchkick/results.rb
@@ -173,21 +173,21 @@ module Searchkick
           end
       end
 
-#      if records.respond_to?(:primary_key) && records.primary_key
-#        # ActiveRecord
-#        records.where(records.primary_key => ids)
-#      elsif records.respond_to?(:all) && records.all.respond_to?(:for_ids)
-#        # Mongoid 2
-#        records.all.for_ids(ids)
-#      elsif records.respond_to?(:queryable)
-#        # Mongoid 3+
-        records.queryable.for_ids(ids)
-#      elsif records.respond_to?(:unscoped) && records.all.respond_to?(:preload)
-#        # Nobrainer
-#        records.unscoped.where(:id.in => ids)
-#      else
-#        raise "Not sure how to load records"
-#      end
+     if records.respond_to?(:queryable)
+      # Mongoid 3+
+      records.queryable.for_ids(ids)
+     elsif records.respond_to?(:primary_key) && records.primary_key
+       # ActiveRecord
+       records.where(records.primary_key => ids)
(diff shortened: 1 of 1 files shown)


ankane:
Hi @benjaminharel, thanks for the pull request!  Can you explain more about what's not working for you, or add a failing test case to demonstrate?

benjaminharel:
Hi @ankane .
I'm working with Mongoid::VERSION => "5.0.0.rc0"
This line of code returns <i>true</i> for mongoid.
`     if records.respond_to?(:primary_key) && records.primary_key
       # ActiveRecord`


So the actual filter in mongoid query looks like:
`"filter"=>{"_id"=>[BSON::ObjectId('564305d1c1c83c0e6f000004'), BSON::ObjectId('5641aab6c1c83c7f3c000002'), BSON::ObjectId('564305abc1c83c0e6f000003')]}`

but it should be
`"filter"=>{"_id"=>{"$in"=>[BSON::ObjectId('564305d1c1c83c0e6f000004'), BSON::ObjectId('5641aab6c1c83c7f3c000002'), BSON::ObjectId('564305abc1c83c0e6f000003')]}}`


So results are found in Elastic, but not fetched from Mongo.

ankane:
Thanks @benjaminharel.  You should be able to fix by upgrading to a stable release of Mongoid.

ankane:
Cleaning up PRs.

lalitlogical:
Hi @ankane, I have below gem versions into my project.

```
mongoid-6.0.3
searchkick-2.4.0
activeadmin-mongoid-0.4.0
```

activeadmin-mongoid define the `primary_key` method. So searching feature is not working because below code where `primary_key` is assumed for **ActiveRecord**. 

```
def self.load_records(records, ids)
    records =
      if records.respond_to?(:primary_key)
        # ActiveRecord
        records.where(records.primary_key => ids) if records.primary_key
      elsif records.respond_to?(:queryable)
        # Mongoid 3+
        records.queryable.for_ids(ids)
      elsif records.respond_to?(:unscoped) && :id.respond_to?(:in)
        # Nobrainer
        records.unscoped.where(:id.in => ids)
      elsif records.respond_to?(:key_column_names)
        records.where(records.key_column_names.first => ids)
      end

    raise Searchkick::Error, "Not sure how to load records" if !records

    records
  end
```


We need to change order of ORM checking condition. Please let me know if I can help.

Thanks

ankane:
Hey @lalitlogical, I don't think it makes sense to update this as a result of an `activeadmin-mongoid` patch. My suggestion would be to add a `where` method to your model so it works with your project.

PR #564, opened 2015-11-26
Reduce default :limit to fix compatibility with Elasticsearch v2.1.0

ledermann:
Elasticsearch v2.1.0 introduced a soft limit on the maximum number of results (default 10,000 - see elastic/elasticsearch#13188). Searchkick uses 100,000 - so without using an explicit limit a query_phase_execution_exception will be raised.

I have changed the Searchkick default to 1000, because this is the recommendation found here:
https://www.elastic.co/guide/en/elasticsearch/guide/current/pagination.html

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -49,7 +49,7 @@ module Searchkick
 
       # pagination
       page = [options[:page].to_i, 1].max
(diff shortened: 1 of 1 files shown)


ankane:
Thanks Georg!  Love rails-settings.

PR #573, opened 2015-12-09
Reverse pin location when passed as array.

klebervirgilio:
Hi there,

I came across this possible bug today. _I wouldn't say bug, it's more for an improvement._

Problem: When sorting via `_geo_distance_` we have the flexibility to pass the lat and lon as a array as well as Hash as we do in the `where` section.

```ruby
{
  :_geo_distance: { location: [-114, 23] } 
  # or
  :_geo_distance: { location: {lat: 23, lon: -114} } 
  # or
  :_geo_distance: { :"pin.location" => {lat: 23, lon: -144} } 
}
```
Right?

My proposal is, as the `where` section takes care of that for us for example [here](https://github.com/ankane/searchkick/blob/master/lib/searchkick/query.rb#L602), [here](https://github.com/ankane/searchkick/blob/master/lib/searchkick/query.rb#L610) and [here](https://github.com/ankane/searchkick/blob/master/lib/searchkick/query.rb#L611) the developer will assume that the same it's being done in the `sort` section.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -705,7 +705,8 @@ module Searchkick
             location = (
               sort[:_geo_distance][:location]  ||
               sort[:_geo_distance][:"pin.location"]
-            )
+            )                                  &&
+            location.is_a?(Array)
 
           location.reverse!
         end
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -371,10 +371,8 @@ module Searchkick
         payload[:explain] = options[:explain] if options[:explain]
 
         # order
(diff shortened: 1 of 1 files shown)


klebervirgilio:
Hey @ankane!
Hopefully [this commit](https://github.com/ankane/searchkick/commit/7a1f2508a4ad72b9d6dc81376c9c21ade5e9719c) will avoid the problem that this commit was suppose to fix. Do you think?

ankane:
Oops, meant to post here yesterday - sorry! This was the final push needed to move to the explicit format. Thanks @klebervirgilio :+1:

PR #577, opened 2015-12-11
add comment on queue used for callbacks: :async

nicolasmlv:


ankane:
Thanks for the suggestion Nicolas!

PR #587, opened 2015-12-17
Use bool filters instead of and/or filters except for geo-spacial, an…

elguapo1611:
…d range filters

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -382,7 +382,9 @@ module Searchkick
         if filters.any?
           if options[:facets] || options[:aggs]
             payload[:filter] = {
-              and: filters
+              bool: {
+                must: filters
+              }
             }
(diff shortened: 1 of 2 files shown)


elguapo1611:
Follows some of the advice from this blog post: https://www.elastic.co/blog/all-about-elasticsearch-filter-bitsets

meetrajesh:
Have we done any tests to prove that this is actually faster for the more recent versions? The blog post linked indicates the advice provided is now deprecated.

elguapo1611:
@meetrajesh this feature is apparently fixed in es 2.0.  And/Or filters will use bitsets, but for everyone still using 1.7, this will perform better.

ankane:
Added the additional test case and ensured there are no more `and` filters. Sorry we couldn't get this in sooner.

PR #591, opened 2015-12-30
Refactor index implementation

st0012:
Hi, I just refactored `index.rb`. Mainly just puts some methods into helper module, or uses decorator pattern to hide some implementations. And I didn't touch any implementation's logic, so I think it should still work well. Hope you can give me your thoughts about this refactoring or any advice (i.e. module or method naming). Thank you :smile:


ankane:
Hey @st0012, thanks for the PR 👍  and sorry for not reviewing it sooner. The index options have since been refactored out.

Issue #196, opened 2014-05-08
Problems Reindexing with Rails 4


mstruve:
I had an [issue](https://github.com/ankane/searchkick/issues/153) before with an outdated multi-json not parsing things correctly but I got that problem resolved with a Rails upgrade.  However now I am running into an elasticsearch problem when I try to reindex.

I am using searchkick version 0.7.4 and Rails 4.0.4.

Model Setup:

```
class Deal < ActiveRecord::Base  
  searchkick

  def search_data
    {
      product_name: product_name,
      brand_name: brand_name
    }
  end
end
```

Trace:

```
aisle50(dev)> Deal.reindex
  Deal Load (52.4ms)  SELECT "deals".* FROM "deals" WHERE "deals"."deleted" = 'f' ORDER BY "deals"."id" ASC LIMIT 1000
Elasticsearch::Transport::Transport::Errors::BadRequest: [400] {"error":"Unexpected end-of-input: expected close marker for OBJECT (from [Source: [B@45dbd25c; line: 1, column: 0])\n at [Source: [B@45dbd25c; line: 1, column: 3]"}
        from /Users/mollystruve/.rvm/gems/ruby-1.9.3-p545@aisle50web/gems/elasticsearch-transport-0.4.11/lib/elasticsearch/transport/transport/base.rb:132:in `__raise_transport_error'
        from /Users/mollystruve/.rvm/gems/ruby-1.9.3-p545@aisle50web/gems/elasticsearch-transport-0.4.11/lib/elasticsearch/transport/transport/base.rb:227:in `perform_request'
        from /Users/mollystruve/.rvm/gems/ruby-1.9.3-p545@aisle50web/gems/elasticsearch-transport-0.4.11/lib/elasticsearch/transport/transport/http/faraday.rb:20:in `perform_request'
        from /Users/mollystruve/.rvm/gems/ruby-1.9.3-p545@aisle50web/gems/elasticsearch-transport-0.4.11/lib/elasticsearch/transport/client.rb:92:in `perform_request'
        from /Users/mollystruve/.rvm/gems/ruby-1.9.3-p545@aisle50web/gems/elasticsearch-api-0.4.11/lib/elasticsearch/api/actions/bulk.rb:81:in `bulk'
        from /Users/mollystruve/.rvm/gems/ruby-1.9.3-p545@aisle50web/gems/searchkick-0.7.4/lib/searchkick/index.rb:44:in `block in import'
[...]

PR #601, opened 2016-01-13
Problems reindexing when using Oj

DanBradbury:
I was using searchkick in a project and had just introduced `Oj` to help optimize and format JSON output. After running my test suite I ran into the following error with a model that was using `reindex`:

`Elasticsearch::Transport::Transport::Errors::BadRequest: [400] {"error":"Unexpected end-of-input.. }`
After further investigation I found this to occur during most other model methods and was not limited to reindex (search, index, callbacks, etc)

It turns out someone else had reported nearly the exact same issue back in May 2014 (#196). In that Issue the OP is content with removing Oj config and have things work as expected.. 
>Turns out the culprit was a single line problem in a Rabl config file that we had put in at one point while trying to debug another situation.
> 
> Oj.default_options = Oj.default_options.merge({indent: 2})
Removed the line, removed the problem. Love a simple solution!

Unfortunately for us removing that config was not an option...

I propose a solution for all Oj  users; which would ensure their options are usable as is by having the `indent` option set to its default state (0) and then switching back to the users settings after processing (and all transport) is completing.

Open to suggestions to the testing as it does seem a bit invasive (introducing `oj` dependency, etc)

### DIFF
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -101,11 +101,16 @@ module Searchkick
   # private
   def self.perform_items(items)
     if items.any?
+      if defined?(Oj)
+        oj_options = Oj.default_options
+        Oj.default_options = { :indent => 0 }
+      end
       response = client.bulk(body: items)
       if response["errors"]
         first_item = response["items"].first
         raise Searchkick::ImportError, (first_item["index"] || first_item["delete"])["error"]
       end
+      Oj.default_options = oj_options if defined?(Oj)
     end
   end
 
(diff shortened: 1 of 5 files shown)


ankane:
Hi Dan, thanks for the pull request! Unfortunately, I don't think this approach is thread-safe. You may be able to write a [custom serializer](https://github.com/elastic/elasticsearch-ruby/tree/master/elasticsearch-transport#serializer-implementations) to strip out the extra whitespace.

PR #605, opened 2016-01-19
add option for number of replicas

neneboe:
Added this option because Heroku Bonsai tiers limit number of shards you can have, and because I may or may not care about failovers.


ankane:
Thanks for the pull request Nick!  Since you can do `searchkick settings: {number_of_replicas: 1}`, I don't think an extra option is needed.

neneboe:
Oh awesome! I was scanning the source code for something like that but must
have missed it. Thanks!

ankane:
Also added a note to the readme. Should have been there in the first place.

Issue #606, opened 2016-01-21
similar method not support per_page option


shiro16:
Hi,
Is this intentional?

use limit option
```
> User.search('*', limit: 2).count
=> 2
> user = User.first
> user.similar(limit: 2).count
=> 2
```

use per_page option
```
> User.search('*', per_page: 2).count
=> 2
> user = User.first
> user.similar(per_page: 2).count
=> 10
```

I expect return 2
I think option unified good.
Thanks.

PR #607, opened 2016-01-21
fix similar method default per_page

shiro16:
see :  similar method not support per_page option #606


ankane:
Great, thanks @shiro16! :+1:

PR #614, opened 2016-01-29
Add example in readme how to search in nested fields

tacnoman:


ankane:
Thanks Renato. Can we add it to the reference section instead? Also, it'd be nice to write a test for this feature if you have time.

ACPK:
This is awesome!

ankane:
There's now a section for nested data: https://github.com/ankane/searchkick#nested-data

PR #616, opened 2016-02-02
adapt to Faraday sending a hash or an object to `call`

### LINKED ISSUE #611: ArgumentError: wrong number of arguments (0 for 1) in env.method
Im getting this error and Im using ruby 1.9.3
The error is in env.method. 'method' function is a function from Kernel class which inherits all objects.
so calling env.method executes Kernel.method instead of the {:method => ""} hash property. thats what I think the problem is.

hmm doing env.class returns Hash
```
ArgumentError: wrong number of arguments (0 for 1)
from /Users/judecalimbas/.rvm/gems/ruby-1.9.3-p484@listing_spark/gems/searchkick-1.1.2/lib/searchkick/middleware.rb:5:in `call'

     6: def call(env)
     8:   if env.method == :get && env.url.path.to_s.end_with?("/_search")
     9:     env.request.timeout = Searchkick.search_timeout
    10:   end
    11:   @app.call(env)
    12: end
```

aaroncraigongithub:
Fixes #611

### DIFF
--- a/lib/searchkick/middleware.rb
+++ b/lib/searchkick/middleware.rb
@@ -3,21 +3,10 @@ require "faraday/middleware"
 module Searchkick
   class Middleware < Faraday::Middleware
     def call(env)
-      is_search = env_value(env, :url).path.to_s.end_with?("/_search")
-
-      if env_value(env, :method) == :get && is_search
-        r = env_value(env, :request)
-        if r.is_a?(Hash)
-          r[:timeout] = Searchkick.search_timeout
-        else
-          r.timeout = Searchkick.search_timeout
-        end
+      if env[:method] == :get && env[:url][:path].to_s.end_with?("/_search")
+        env[:request][:timeout] = Searchkick.search_timeout
       end
       @app.call(env)
-    end
-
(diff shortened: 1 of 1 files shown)


ankane:
Hey Aaron, thanks for the pull request. What version of Faraday are you on that expects a hash?

aaroncraigongithub:
```
$ bundle list | grep faraday
  * faraday (0.8.11)
  * faraday_middleware (0.9.0)
```

ankane:
Looks like the hash approach `env[:url]`, etc can be used with both older and newer versions of Faraday.

aaroncraigongithub:
I did not check that!  I can make that adjustment and update this PR, if that works for you.

ankane:
That'd be great. Should simplify things.

aaroncraigongithub:
I believe this is good to go now, though I haven't tested on any other version of Faraday.

ankane:
Tested with `0.8.11` and `0.9.2`.  Should be good.

ankane:
Thanks :+1:

PR #619, opened 2016-02-04
Allow new way of selecting source fields

meetrajesh:
**Problem**: The old way of selecting a subset of source fields was to use `select:` in the search options which maps to `fields:` in the ES query body. This had two major problems:

1. You can only select leaf nodes
2. Results were returned as an array regardless of how they're originally stored in the document (*really* annoying)

The new way of selecting subset of source fields is to use the `_source` option in the search payload: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-source-filtering.html

The new way fixes the limitations of the old way, while adding a few new cool features:

1. You can now select non-leaf-nodes
2. You can now add wildcards within the object tree, eg. `obj.*`, `obj1.*`, etc. The old feature only operated on leaf nodes so this wasn't needed.
3. You can specify both `include` and `exclude` patterns. Previously you could only specify a whitelist (include), but not a blacklist (exclude). This allows a user to select all fields _except_ 3 specified fields, for example.

This PR introduces a new option called `source_select` that behaves mostly like `select` but uses the new ES source selection implementation.

The only difference between this PR's implementation and the documented `_source:` feature is that I've changed `source_select: []` to return no source fields to match the existing `select: []` behavior. The default elasticsearch behavior for `_source: []` is to return _all_ fields, similar to `select: true`, or no select at all.

`source_select: true` returns all fields. This is the default ES behavior of `_source: true`.

### DIFF
--- a/lib/searchkick/model.rb
+++ b/lib/searchkick/model.rb
@@ -73,7 +73,7 @@ module Searchkick
         callback_name = callbacks == :async ? :reindex_async : :reindex
         if respond_to?(:after_commit)
           after_commit callback_name, if: proc { self.class.search_callbacks? }
-        else
+        elsif respond_to?(:after_save)
           after_save callback_name, if: proc { self.class.search_callbacks? }
           after_destroy callback_name, if: proc { self.class.search_callbacks? }
         end
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -567,10 +567,19 @@ module Searchkick
         end
 
         # An empty array will cause only the _id and _type for each hit to be returned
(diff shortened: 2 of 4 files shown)


ankane:
PR looks great. This should def be the default in Searchkick 2.0. Want to think on the name for a bit.  Alternatives include `select_v2`, `select2`, `select_source`, and `sel`.  Leaning towards the first.

meetrajesh:
@ankane `select_v2` sounds good, especially if it's going to be the default in Searchkick 2.0.

ankane:
Cool, will go with that. Merged! :+1: Not sure why Github didn't recognize it

PR #624, opened 2016-02-09
Refactored Searchkick::Query#prepare

doubiz:
Refactored Searchkick::Query#prepare. Separated logic in methods


ankane:
Great, thanks @doubiz! :+1:

PR #625, opened 2016-02-10
Recommend touch: true for "cache-busting"

aimee-ault:
The README here suggests using ActiveRecord callbacks to trigger a reindex on an association update. It seems like it'd be much cleaner if you just used the `touch` option, which updates the parent record's `updated_at` timestamp when the child object is changed.

P.S. HUGE fan of this gem. Thank you for authoring it!

### DIFF
--- a/README.md
+++ b/README.md
@@ -428,17 +428,11 @@ end
 
 #### Associations
 
-Data is **not** automatically synced when an association is updated.  If this is desired, add a callback to reindex:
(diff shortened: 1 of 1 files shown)


ankane:
Hi Aimee, thanks for the pull request! It looks like `touch: true` will work as of ActiveRecord 4.0, but I'm a bit hesitant to make the change since some people may not always want to modify `updated_at` and it doesn't work for Mongoid (`touch: true` is a valid option, but it doesn't run callbacks, same as ActiveRecord before 4.0).

ankane:
Closing for the concerns above.

PR #626, opened 2016-02-11
Add require statement to signed request example in README

abevoelker:
Without this require, I was getting

```
/var/bundle/ruby/2.3.0/gems/faraday-0.9.2/lib/faraday.rb:189:in `lookup_middleware': :aws_signers_v4 is not registered on Faraday::Request (Faraday::Error)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -964,6 +964,7 @@ gem 'faraday_middleware-aws-signers-v4'
 and add to your initializer:
 
 ```ruby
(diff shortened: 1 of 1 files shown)


ankane:
Great, thanks Abe for this and the many times you've helped others with issues. :+1:

abevoelker:
You're very welcome! Although I don't think I've helped out very much. Keep up the great work, love searchkick and the rest of your gems!

PR #631, opened 2016-02-17
Raise with message on bulk import error

### LINKED ISSUE #630: Bulk import failure raises Searchkick::ImportError without a message
Hi there,

I noticed that Searchkick prints nothing but `Searchkick::ImportError: Searchkick::ImportError` with a rather unhelpful stack trace when bulk importing fails.

The problem is that Searchkick is looking at only the first item returned for the error message, so when the first item is okay but any other items in the batch aren't, ImportError is raised with a nil message. I'll open a PR with a fix.

Thanks for writing this awesome gem--it's saved me a ton of time :)

audreyschwarz:
Fixes #630. I've implemented it to print every error message and id that comes back, which seemed the most sensible option. Let me know if anything needs changing, and I hope this is helpful!

### DIFF
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -103,8 +103,12 @@ module Searchkick
     if items.any?
       response = client.bulk(body: items)
       if response["errors"]
-        first_item = response["items"].first
-        raise Searchkick::ImportError, (first_item["index"] || first_item["delete"])["error"]
+        messages = response["items"].map do |item|
+          (item["index"] || item["delete"])
+        end.select { |item| item["error"] }.map do |item|
+          "#{item["error"]} on item with id '#{item["_id"]}'"
+        end.join "\n"
+        raise Searchkick::ImportError, messages
       end
     end
   end
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -103,12 +103,10 @@ module Searchkick
     if items.any?
       response = client.bulk(body: items)
       if response["errors"]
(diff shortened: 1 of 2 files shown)


ankane:
Hi @audreyschwarz , thanks for the pull request!  Here are my current thoughts:

- To prevent really long error message, I think we should just show the first error
- It looks like tests are failing for Mongoid 2, 3, and 5. You can test locally with `BUNDLE_GEMFILE=test/gemfiles/mongoid5.gemfile bundle exec rake`
- Try `assert_raises` instead of rescuing the exception

audreyschwarz:
Hm... I could consistently reproduce the failures with ES 1.7.3, but this change fixed that locally. I'll mess around some more and see if I can figure out what's going on.

audreyschwarz:
Lol nevermind, I can reproduce these failures.

tilsammans:
Having this merged would be very helpful @ankane :smile: would it be good to go?

tilsammans:
When this is merged I would like to propose a new patch. I'd like the option to not `raise` when there are errors but print the error in question to STDOUT. This will allow us to import all records besides the one(s) with problematic values.

hooopo:
@ankane This is really useful!

ankane:
Great, thanks @audreyschwarz!

@tilsammans all records in the batch that don't error *are* imported (before the error is raised), so you can simply rescue the error and keep going.

tilsammans:
@ankane I realized that too after I read the source. 👍

PR #635, opened 2016-02-18
Add Support for min_doc_count Filter in Aggregations Query with Tests

aengesath:
Added support for this very useful ES feature: 

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html#_minimum_document_count_3

It allows users to change the minimum number of hits required for results to be returned in aggregations. 

Why is it important? Consider this SQL to find duplicate names:

`select count(*) from products group by name having count(*) >= 2`

With this PR you can now do this:

`Product.search "*", aggs: {name: {min_doc_count: 2}}`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ tmp
 *.log
 .DS_Store
 .ruby-*
+.idea/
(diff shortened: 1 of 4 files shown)


ankane:
Tests and docs... great PR!  Thanks Adam :100:

PR #639, opened 2016-02-29
[WIP] Records order preservation ActiveRecord POC

### LINKED ISSUE #463: How to get ids of results items?
Because this approach smells bad and I have to do two queries to the database.

``` ruby
ids = Model.search(where: where_params)
  .results.response['hits']['hits'].map{ |x| x['_id'].to_i }
```

Is there proper way for doing this?

By the way, I need items ids for "converting" `Searchkick::Results` class to `ActiveRecord::Relation` to use scopes and other class methods on it. Is there turnkey solution for these needs?

``` ruby
...

def self.filter(params)
  where_params = Search.attrs(params: params)

  if where_params.any?
    to_ar_relation(search(where: where_params))
  else
    where(nil)
  end
end

private

# converts searchkick results to ActiveRecord relation to use scopes on it
def self.to_ar_relation(results)
  where(id: results.response['hits']['hits'].map{ |x| x['_id'].to_i })
end

...
```

Mavvie:
Fixes #463 

As you know, currently the `records` method is not ordered by Elasticsearch's `hit` ordering. Taking heavy inspiration from `order_as_specified` gem, this preserves the order of the search results.

It should be noted that this is _quite_ slow to do, and should only be done if you **really** need an `ActiveRecord::Relation`. Otherwise, it's possibly better to just use the existing in-memory representation.
#### Todo
- [ ] Ensure compatibility with all ActiveRecord DBs (MySQL, PostgreSQL, SQLite, etc)
- [ ] Ensure that the existing `results_query` preserves order for Mongoid/NoBrainer

Let me know if you have any thoughts.

also, cc @abevoelker @milushov

### DIFF
--- a/lib/searchkick/results.rb
+++ b/lib/searchkick/results.rb
@@ -17,10 +17,15 @@ module Searchkick
 
     # experimental: may not make next release
     def records
-      @order_by ||= hits.map do |hit|
-        "#{klass.table_name}.#{klass.primary_key}='#{hit['_id']}' DESC"
-      end.join(", ")
-      @records ||= results_query(klass, hits).reorder(@order_by)
+      @records ||= results_query(klass, hits)
+      # Handle order preservation in ActiveRecord
+      if @records.respond_to?(:primary_key) && @records.primary_key
+        @order_by ||= hits.map do |hit|
+          "#{klass.table_name}.#{klass.primary_key}='#{hit['_id']}' DESC"
+        end.join(", ")
+        @records = @records.reorder(@order_by)
+      end
+      @records
     end
 
     def results
--- a/lib/searchkick/results.rb
+++ b/lib/searchkick/results.rb
@@ -17,7 +17,10 @@ module Searchkick
 
     # experimental: may not make next release
     def records
-      @records ||= results_query(klass, hits)
+      @order_by ||= hits.map do |hit|
+        "#{klass.table_name}.#{klass.primary_key}='#{hit['_id']}' DESC"
+      end.join(", ")
+      @records ||= results_query(klass, hits).reorder(@order_by)
     end
 
     def results
--- a/test/records_test.rb
+++ b/test/records_test.rb
@@ -1,8 +1,9 @@
 require_relative "test_helper"
 
 class RecordsTest < Minitest::Test
-  def test_records
-    store_names ["Milk", "Apple"]
-    assert_equal Product.search("milk").records.where(name: "Milk").count, 1
+  def test_records_preserves_order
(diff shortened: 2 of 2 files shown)


jimmybaker:
I've also made a pull request to solve this problem using a different and more performant way. https://github.com/ankane/searchkick/pull/672

Of course milage may vary but I was using a method similar to one you are proposing here and the query time to bring by 25 records was ~100ms which is slow as you mentioned in the description. With what I'm proposing in my PR I was able to get the same 25 records in 0.7ms.

image 2016-05-06 at 1 23 44 pm

Check it out and let me know if you have any thoughts on it.

ankane:
Hey @Mavvie, thanks for the PR. Here's how Elasticsearch Rails handles it.

- [ActiveRecord](https://github.com/elastic/elasticsearch-rails/blob/15761247f3e99654bda946a178e50b5365414b59/elasticsearch-model/lib/elasticsearch/model/adapters/active_record.rb#L28)
- [Mongoid](https://github.com/elastic/elasticsearch-rails/blob/15761247f3e99654bda946a178e50b5365414b59/elasticsearch-model/lib/elasticsearch/model/adapters/mongoid.rb#L22)

It should be a more performant approach.

jimmybaker:
@ankane Has this been done in searchkick yet?

jimmybaker:
@ankane - Would the way ElasticSearch Rails handles it work for us when we paginate search results?

jimmybaker:
This is an interesting problem. I'm working up a PR now and when writing the test I noticed that calling normal enumerable methods like `map`, `each`, or `collect` on `.records` doesn't call `to_a`. Should we also override these methods to ensure that things are done in the proper order?

jimmybaker:
Since we're returning an active record relation with `.records` I still think the best way to handle this is to ensure the ordering is done on the db side based on the ordering of hits from es.

Anyone else following me here?

Mavvie:
I think something similar is possible, but I'm not sure exactly how it would be implemented. I don't really know how pagination libraries work. 

It could make sense to do the ordering in Ruby-land if we're only loading `per_page` records from the db. However, the db might still do it better in that case as well. Not sure really.

khiav223577:
Maybe you could try the order method le0pard shared here (no need to define a db function): https://github.com/khiav223577/find_with_order/issues/5

image

Mavvie:
@ankane maybe we could add that gem as a dependency? Or should we copy what we need from it?

ankane:
Hey, the plan is to remove the `records` method in Searchkick 3.0, so going to close this PR. Feel free to create another gem with this functionality that can tie into Searchkick.

PR #648, opened 2016-03-16
Add min_score option and tests

abookyun:
Hi @ankane 

It's been a while that PRs related to `min_score`(https://github.com/ankane/searchkick/pull/292, https://github.com/ankane/searchkick/pull/492) not being merged. I guess that's because the deprecated facets and missing tests makes you labeled `waiting`, so I just extracted `min_score` option part from PRs and had rebased master already. Thanks to @sebfie and @st0012, hope this small feature could be merged.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -158,6 +158,7 @@ module Searchkick
       per_page = (options[:limit] || options[:per_page] || 1_000).to_i
       padding = [options[:padding].to_i, 0].max
       offset = options[:offset] || (page - 1) * per_page + padding
+      min_score = options[:min_score] || 0.0
 
       # model and eagar loading
       load = options[:load].nil? ? true : options[:load]
@@ -342,7 +343,8 @@ module Searchkick
         payload = {
           query: payload,
           size: per_page,
-          from: offset
+          from: offset,
+          min_score: min_score
         }
         payload[:explain] = options[:explain] if options[:explain]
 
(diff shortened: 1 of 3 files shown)


hendricius:
👍 this would be great to have.

zph:
@abookyun Thanks for posting this PR. I pulled it onto a branch and it seems to be working well for my testing.

ankane:
Thanks David, it'd be great to have a more general option for this (one that could merge any code with the body).  Ideally we could accomplish this with the `body` option (as people have expected this based on past questions), but that'd break backwards compatibility right now.  Maybe something like:

```ruby
Product.search("*", merge_body: {min_score: 1})
```

I'm not in love with the name.  Let me know your thoughts.

abookyun:
How about `body_options`?

```ruby
Product.search("*", body_options: { min_score: 1.0 })
```

Then merge `options[:body_options]` with `options[:body]` right before original preparing process. That could accomplish `a more general option for this (one that could merge any code with the body)` and if there's any naming conflict mess up with original process, the original process overrides them which shouldn't happen. For example:

https://github.com/ankane/searchkick/blob/master/lib/searchkick/query.rb#L170

```ruby
options[:body].merge(options[:body_options]) if options[:body_options]
options[:json] ||= options[:body]

# or with lonely operator
options[:json] ||= options[:body]&.merge(options[:body_options])
```

ankane:
Let's go with `body_options`.  I think we want to apply it after the payload is constructed, possibly here: https://github.com/ankane/searchkick/blob/master/lib/searchkick/query.rb#L395

ankane:
Also, we need to decide if it should be a shallow merge or a deep merge.  Unclear what other use cases will be, but leaning towards deep.

abookyun:
Patch a commit that added new option `:body_options` and make it deep merged with body after payload was constructed.

Should patch https://github.com/ankane/searchkick#advanced-search with body_options will be deep merged into body cautions? /cc @ankane

ankane:
Looks good. I'll add it to the reference section and cut a new release.

ankane:
Thanks @abookyun 👍

PR #651, opened 2016-03-23
Async reindex jobs have a potential race condition

neojin:
Hello,

If you have 2 records that get updated in rapid succession and you run the searchkick async reindexer, there is a small chance that the reindex uses old data. Example:
1. record gets updated `name: 'foo'`
2. async reindex gets queued and job runs, but there is some sort of network delay
3. record gets updated `name: 'bar'`
4. async reindex gets queued and job runs and completes _first_
5. job from (2) completes _second_ and now the record's index has `name: 'foo'`, when it should be `name: 'bar'`

To prevent this, I wrote a "queue" with a mutex which counts the number of active jobs that want to index the same record (identified by `"#{klass}::{id}"`). The queue itself contains Mutex instances such that when many jobs are acting on a single record, it basically runs them serial to prevent the above situation.

I don't know if this is the best solution, so looking for feedback and ideas. If you accept this PR I will write the tests for it. Speaking of tests, is there a special command to run `ReindexV2JobTest`? I know it's because `ActiveJob` isn't defined, but not sure how to run it correctly.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright (c) 2016 Andrew Kane
+Copyright (c) 2013-2016 Andrew Kane
 
 MIT License
 


ankane:
That for the pull request Jin!  The race condition still exists if you have multiple worker processes (or multiple web processes enqueuing jobs), so I'm not sure this accomplishes its goal in most production situations.

neojin:
Hey Andrew, thanks for the feedback. Man, I totally forgot about the multiple worker processes, good call. We're only running a single process with 20 workers and completely overlooked that possibility. I have been using sidekiq/redis with ActiveJob, and again totally forgot that you can run ActiveJob inline.

In light of this new information, do you have any ideas on how to proceed? I was thinking about doing something with a temp file and `flock`, but that wouldn't work since you can run the processes on different boxes. They I thought about doing something with redis, but that wouldn't work since ActiveJob can be run in memory.

I am very motivated to provide a solution for this as I am running into this problem daily and have been forced to run a serial reindex queue with a single worker to keep things in sync, but it is getting frequently backed up. Any ideas would be greatly appreciated.

fabn:
Relevant article: http://lanceolsen.net/preventing-race-conditions-with-fencing-tokens-in-rails/

bsodmike:
This may be an aside, the scenario I'm looking at the moment is this:

- ELB (load-balancer at point of ingress)
- Traffic is routed to 1 of N nodes (instances).
- Each instance runs an app instance as well as a single worker (hypothetically, bare with me).

Suppose traffic is routed to instance 1, and a user deletes a post.  And the next request is routed to instance 2 and someone has just posted an update to a post.

Let's now assume the delete job fires before the update, updating the ES index.  💣 
Thoughts?

ankane:
Cleaning up PRs. As far as I know, a distributed transaction is the only way to keep things 100% in sync, which is outside the scope of the gem. Thanks again for the PR @neojin 👍

PR #662, opened 2016-04-24
Avoid extra string allocations

estum:
Avoid extra string allocations on [Index#document_type(record)](/ankane/searchkick/blob/master/lib/searchkick/index.rb#L538), which by default fallbacks to `model_name.to_s.underscore` which creates a new string object and call Inflector every time it used. Models already have `model_name.singular`, which is frozen, we should you it to skip a generation of tons of strings inside a huge reindexing batch.


ankane:
Great, thanks @estum!

PR #672, opened 2016-05-05
Maintain ordering of records from ES response

jimmybaker:
I questioned whether or not I should submit a PR for this, but it seems like there are lots of people having the same issue where the records you get from Searchkick::Results are not in the same order that ElasticSearch responded with.

Based on my small amount of testing, the solution is performant and the only drawback is that you must define a db function if you're using Postgres. I've included an example postgresql function in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,30 @@ order: {_score: :desc} # most relevant first - default
 
 [All of these sort options are supported](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-sort.html)
 
+Using PostgreSQL? Are your records out of order?
+
+When requesting records from a postgresql database, the records are not in the same order that they were returned from ElasticSearch. In order to make this work and perform well, you will need to implement a sorting function at the database level. This is an example function in postgresql:
+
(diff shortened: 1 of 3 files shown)


Mavvie:
Hmm, I do like the approach of using a postgresql function instead of what I did in #639, but I'm not 100% sure this should be merged as-is. I'm not sure how stored functions work, but I don't like the pattern of "paste this in your SQL console". It becomes easy to forget to do this in all your different environments, etc., and next thing you know your production env is broken because you forgot to install the function.

Perhaps changing the `README` to include it as a migration would make more sense? If it's possible to do this as a migration, or add some failsafe in case the postgresql function isn't defined at runtime, then I 👍 this approach.

jimmybaker:
@Mavvie Yeah I like the migration idea. I'll update the pr soon. Thanks!

ankane:
Hey @jimmybaker, thanks for the PR 👍  I think a cleaner approach would be this: https://github.com/ankane/searchkick/pull/639#issuecomment-259072405

PR #708, opened 2016-07-29
Support conversion boosting on multiple fields

meetrajesh:
We want to be able to a/b test different scoring mechanisms for conversion boosting. Right now it's not possible to specify the conversions field dynamically at query-time. It can only be specified in the options hash when initializing Searchkick.

This PR changes that and allows the query to overwrite the conversions field that is used to boost records.

```
class Product < ActiveRecord::Base
  has_many :searches, class_name: "Searchjoy::Search"

  searchkick conversions: ["unique_user_conversions", "total_conversions"]

  def search_data
    {
      name: name,
      unique_user_conversions: searches.group(:query).uniq.count(:user_id)
      # {"ice cream" => 234, "chocolate" => 67, "cream" => 2}
      total_conversions: searches.group(:query).count
      # {"ice cream" => 412, "chocolate" => 117, "cream" => 6}
    }
  end
end
```

At query time, you can choose to boost by both fields (default), just one of them, or none of them:

```
Product.search("banana", {conversions: "total_conversions"}) # only boost by total_conversions
Product.search("banana", {conversions: false}) # no conversion boosting
```

cc: @ankane

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -162,8 +162,8 @@ module Searchkick
       # model and eagar loading
       load = options[:load].nil? ? true : options[:load]
 
-      conversions_field = searchkick_options[:conversions]
-      personalize_field = searchkick_options[:personalize]
+      conversions_fields = Array(options[:conversions] || searchkick_options[:conversions])
+      personalize_field  = searchkick_options[:personalize]
 
       all = term == "*"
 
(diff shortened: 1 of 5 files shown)


lib/searchkick/query.rb
```diff
@@ -275,35 +275,62 @@ def prepare
...
+                                  }
+                                }.merge(script_score)
+                              }
+                            }
                           }
-                        }
-                      }.merge(script_score)
+
+
+              payload = {
```
ankane: @meetrajesh what's the reason for changing the payload here vs adding the shoulds below?

meetrajesh: @ankane this is the old code that i forgot to delete. i was using it as a reference to write the new code but forgot to clean up. thanks for catching!

ankane:
@meetrajesh feel free to skip the test if `elasticsearch_below14?`

ankane:
Even 1.4 was just reached its end for support. https://www.elastic.co/support/eol

ankane:
Great, thanks @meetrajesh 💯

PR #722, opened 2016-09-02
Update readme to fix Searchjoy relation

kevinkimball:
Since this relationship is polymorphic, need to specify `as: :convertable` or Rails throws an error.


ankane:
Great, thanks @kevinkimball 👍

PR #723, opened 2016-09-04
Silent test warnings

savayg:
When I ran the tests I was flawed with warnings, mainly because of Elasticsearch gem not setting the content-type header.

### DIFF
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -40,7 +40,7 @@ module Searchkick
     @client ||=
       Elasticsearch::Client.new(
         url: ENV["ELASTICSEARCH_URL"],
-        transport_options: {request: {timeout: timeout}}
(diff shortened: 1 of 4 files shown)


ankane:
Thanks @savayg! I'm not a huge fan of the `defined?` pattern. Can we set those values to nil in the constructor instead?

nandokakimoto:
Hi @savayg and @ankane. I am wondering if we want to [setup headers](https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch-transport/test/integration/client_test.rb#L73) or just silent test warnings?

savayg:
@nandokakimoto thanks for pointing that out, much better now. @ankane I agree, just updated the PR

ankane:
Very clean, thanks @savayg and @nandokakimoto! 💯

Issue #731, opened 2016-09-15
Directional Synonyms


MK-42:
according to [Elasticsearch: The Definitive Guide](https://www.elastic.co/guide/en/elasticsearch/guide/current/synonym-formats.html) it is possible to define directional synonyms that are only expanded in one direction.

Is it possible to define such synonyms with searchkick? If yes, how to do that?

I looked into the indexing-code if searchkick and found

```
settings[:analysis][:filter][:searchkick_synonym] = {
   type: "synonym",
   synonyms: synonyms.select { |s| s.size > 1 }.map { |s| s.join(",") }
}
```

what in my understanding is the part that creates the synonym-option for elasticsearch. I found no other references to synonym creation in the code.

Would you like to see the `=>` syntax added to searchkick? If yes I could try to integrate that.

Edit:
as a hotfix, I just removed the `select { |s| s.size > 1 }` and passed the full string "foo => bar,baz" into searchkick

ankane:
@MK-42 I'm happy to accept a PR that allows you to pass an array of strings to `synomyms`.

MK-42:
I just found your
```
# TODO use a snowball stemmer on synonyms when creating the token filter
```
what is actually also something i would need.
I tried stemming with [ruby-stemmer](https://github.com/aurelian/ruby-stemmer), but that gave different stemmings than elasticsearch. Is there any documentation on the stemmer used by elasticsearch?

PR #733, opened 2016-09-16
Added directional synonyms with '=>' syntax (#731)

MK-42:
Fixes Issue #731


ankane:
Thanks for putting this together @MK-42. I think the format for directional synonyms should be a string rather than an array of strings with a single element.

oyeanuj:
@MK-42 is the change as per @ankane's suggestion still on your radar? This would be a very useful feature to have!

ankane:
Hey @oyeanuj, I added this just now in a62efa364c6eb5af834022eeae8087a115593914. Thanks for the initial work @MK-42 👍

PR #739, opened 2016-09-27
Added `Searchkick::Model#searchkick_indexed_fields` to get which fields are indexed for a model

oss92:
Imagine you are in you controller and want to get the fields to list in a dropdown for example. There was no way to achieve that.


ankane:
Hey @oss92, thanks for the PR! My initial thoughts is the use case probably isn't super common, and since it's easy enough to do `Model.new.search_data.keys`, there isn't a strong reason to include it in the gem.

PR #752, opened 2016-10-19
Add support for geo_polygon

dwightwatson:
This simply adds support for the `geo_polygon` option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -852,6 +852,12 @@ Bounded by a box
 City.search "san", where: {location: {top_left: {lat: 38, lon: -123}, bottom_right: {lat: 37, lon: -122}}}
 ```
 
+Bounded by a polygon
+
+```ruby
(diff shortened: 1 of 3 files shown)


ankane:
Thanks @dwightwatson, can you add a test case for it and a short example in the readme? Existing [tests for geo filters are here](https://github.com/ankane/searchkick/blob/master/test/where_test.rb#L106) and [docs are here](https://github.com/ankane/searchkick#geospatial-searches).

dwightwatson:
Hey @ankane, I've added a test case and documented it.

One thought though, instead of `geo_polygon` would it be better to name it `within`? That way the syntax would look like `within: {points: x}`. I do notice you currently have `within` stubbed out and was wondering if there was a reason for it, or if this would be a good opportunity to use it.

```
when :within, :bottom_right
  # do nothing
```

ankane:
Awesome, looks great! 💯  `within` is used as part of the `near` option (used in a different part of that code)

Issue #740, opened 2016-09-28
Autocomplete section of README.md is outdated


mimikadze:
Seems that typeahead.js version moved far ahead and they've changed syntax and set of params.

Current solution they provided and I can make it work looks like this:

```
<input class="typeahead" type="text" placeholder="Choose the book">

<script src="https://code.jquery.com/jquery-3.1.1.min.js"></script>
<script src="https://twitter.github.io/typeahead.js/releases/latest/typeahead.bundle.js"></script>

<script>
  var books = new Bloodhound({
    datumTokenizer: Bloodhound.tokenizers.obj.whitespace('title'),
    queryTokenizer: Bloodhound.tokenizers.whitespace,
    remote: {
      url: '/books/autocomplete?query=%QUERY',
      wildcard: '%QUERY'
    }
  });
  $('.typeahead').typeahead(null, {
    source: books
  });
</script>
```

Works for typeahead.js 0.11.1 which is latest for the moment.

I'm sure it makes sense cause my colleagues stuck with this issue recently. I can make a PR if you consider this useful.

ankane:
Thanks for reporting @mimikadze, a PR would be great.

natedrouin:
This is correct, very outdated.

PR #759, opened 2016-10-31
Autocomplete section of README.md updated

mimikadze:
Late update on your request https://github.com/ankane/searchkick/issues/740#issuecomment-250227519


ankane:
Hey @mimikadze, thanks for the PR 👍 

Can you make the changeset smaller? We should only need to update what's in between the `<script>` tags.

ankane:
I went ahead and pushed a more minimal update. Thanks for kicking this off.

PR #768, opened 2016-11-16
minor PR cleaning up redundancy in readme

josh-m-sharpe:
..because of this:

https://github.com/ankane/searchkick/blob/979ae68e8634cb053a717925118af3dc4ec64b5d/lib/searchkick.rb#L42

Though, I'd argue if you're going to use Typhoeus just go full-bore and require the thing.  Is there a reason you're hedging?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1048,7 +1048,7 @@ rake searchkick:reindex CLASS=Product
 
 ### Performance
 
-For the best performance, add [Typhoeus](https://github.com/typhoeus/typhoeus) to your Gemfile.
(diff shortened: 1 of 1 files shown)


ankane:
Hey @josh-m-sharpe, thanks for the PR 👍   It makes sense to remove the extra line from the initializer, but I don't see a strong reason to change the wording.

josh-m-sharpe:
That's fair, any reason not to get rid of all of this and add typhoeus as a dependency?  Seems like an overcomplicated setup process if this addition makes things better.

ankane:
I'm a big fan of simple is better, but It's generally a good practice to keep dependencies to a minimum. I think that wins out here, as the set up isn't terribly complicated.

ankane:
The readme now reflects this.

PR #771, opened 2016-11-19
Add partial_reindex instance method

paddingtonsbear:
Reindex subsets of a particular record
```ruby
class Product
  def search_pricing
    {
      price: price
    }
  end
end
#
product = Product.first
product.partial_reindex(:search_pricing)
```


ankane:
Great, thanks @paddingtonsbear 👍  Made small changes to the test to get it working with ActiveRecord below 4 and Mongoid.

Issue #728, opened 2016-09-12
Avoiding IllegalArgumentException where utf > 32766


mrjlynch:
Solutions for the `IllegalArgumentException` where text is longer than 32766 characters is discussed on [SO](http://stackoverflow.com/questions/24019868/utf8-encoding-is-longer-than-the-max-length-32766). 

One such [solution](http://stackoverflow.com/questions/24019868/utf8-encoding-is-longer-than-the-max-length-32766#answer-30522942) provides three options:
- set "index": "no"
- set "index": "analyze"
- set "index": "not_analyzed" and "ignore_above": 256

How can I set this using searchkick? Is the following correct?

``` ruby
searchkick mappings: {
    product: {
      properties: {
        name: {type: "string", index: "analzye", ignore_above: 256}
      }
    }
  }
```

ankane:
The latest version of Searchkick should set this automatically if you're on Elasticsearch 2.2 or above. You may need to reindex to update the mapping if upgrading from an earlier version of either project.

https://github.com/ankane/searchkick/blob/master/lib/searchkick/index.rb#L551

mrjlynch:
Thanks @ankane, when you say Searchkick sets this automatically, do you mean that default behaviour permits insertion where UTF > 32766?

ankane:
It uses the `ignore_above` option for it.

mrjlynch:
I am using `2.2.1` and I am still getting this error: 

``` ruby
Searchkick::ImportError - {"type"=>"illegal_argument_exception", "reason"=>"Document contains at least one immense term in field=\"html\" (whose UTF8 encoding is longer than the max length 32766), all of which were skipped.  Please correct the analyzer to not produce such terms.  The prefix of the first immense term is: '[60, 100, 105, 118, 62, 68, 105, 97, 103, 110, 111, 115, 105, 115, 32, 111, 102, 32, 99, 101, 108, 105, 97, 99, 32, 100, 105, 115, 101, 97]...', original message: bytes can be at most 32766 in length; got 47021", "caused_by"=>{"type"=>"max_bytes_length_exceeded_exception", "reason"=>"bytes can be at most 32766 in length; got 47021"}}
```

Do I need to set anything anything else to avoid this error?

mrjlynch:
My error, I forgot to `reindex` my dad

mrjlynch:
Is there a way to customize the `ignore_above` setting?

256 is quite short and is no where close to 32766

[...]

PR #773, opened 2016-11-22
Add option to customise ignore_above

mrjlynch:
As discussed in #728, `256` charters is quite restrictive and falls far short of the max of `32766` imposed by Elasticsearch. 

This commit adds the ability to customize the 256 characters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1116,14 +1116,6 @@ class Product < ActiveRecord::Base
 end
 ```
 
-By default, if you are using Elasticsearch 2.2 or above fields above `256` charaters are ignored in `where` statements and exact matches. To customize this settings modify the `ignore_above` setting:
-
-```ruby
-class Product < ActiveRecord::Base
-  searchkick ignore_above: 1000 # max = 32766
-end
-```
-
 ### Advanced Search
 
 And use the `body` option to search:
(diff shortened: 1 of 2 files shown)


ankane:
Thanks @mrjlynch! 👍  Code looks good. As is, the readme update seems a bit confusing/inaccurate. The limit only applies to `where` statements and exact matches.

mrjlynch:
Thanks @ankane, readme is updated. Does that make is less confusing? cheers

mrjlynch:
Would you like me to further review the readme? Thanks @ankane

ankane:
I'm still thinking this will lead to more confusion since it's most likely only relevant for a very small percentage of users. I'm happy to merge the code change, but think we should go without a readme update.

mrjlynch:
Hi @ankane I have updated the readme and removed the readme.

ankane:
Great, thanks @mrjlynch 💯

feliperaul:
@ankane First, congrats for this amazing project. I love searchkick.

As you can probably guess from #850, I spent an entire morning trying to understand why the exact same query would return hits only if a field was specified. 

Looking at the source code I know think it's due to `ignore_above` param, which is set to `256` as default. 

To be honest, I think readme should reflect this.

PR #787, opened 2016-12-19
Handle update errors

klebervirgilio:
Hi,

I recently had an `NoMethodError: undefined method '[]' for nil:NilClass` error being raised when trying to update a document.  After some digging, I noticed searchkick wasn't mapping update errors. 

Should it?


ankane:
Yeah, it should. Nice catch @klebervirgilio. I think it should be `item` instead of `items`.

klebervirgilio:
Cool! I just changed and rebased.

ankane:
Great, thanks!

PR #808, opened 2017-01-12
Allow body to be specified for aggs to enable advanced aggregations

roybotnik:
I didn't see a way to do this right now. Currently it looks like you can only use a few types of aggregations or else it will simply remap to a default bucket terms aggregation. This PR allows for an aggregation body to be specified (got the idea since body is used for advanced searching). This lets us use all of elasticsearch's aggregations without having code for each one:

```
Product.search("*", aggs: {average_price: {body: {avg: {field: :price}}}})

Product.search("*", aggs: {price_stats: {body: {stats: {field: :price}}}})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -710,6 +710,13 @@ Date histogram
 Product.search "pear", aggs: {products_per_year: {date_histogram: {field: :created_at, interval: :year}}}
 ```
 
+Metrics aggregations and more! Specify a `body` to provide aggregation params manually.
+For example, to get an [average](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-avg-aggregation.html) for a particular field:
+
+```
+Product.search "pear", aggs: {average_price: {body: {avg: {field: :price}}}}
+```
+
 #### Moving From Facets
 
 1. Replace `facets` with `aggs` in searches. **Note:** Stats facets are not supported at this time.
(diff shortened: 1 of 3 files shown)


ankane:
Hey @roybotnik, thanks for the PR 👍  You should be able to accomplish this with:

```ruby
Product.search "pear", body_options: {aggs: {average_price: {avg: {field: :price}}}
```

I like this approach better since it doesn't require an additional option and doesn't mix Searchkick syntax and ES syntax. Happy to add the docs section - think people would find it useful.

ankane:
Think the test would be good as well.

roybotnik:
Sure, that looks good, thanks for the tip 😄 . Adding something to the docs would be good because I obviously didn't figure it out 😅 . I was just looking for a quick way to add some support, so if it already works that's perfect. 

Didn't have time yesterday, but when I get a chance I'd like to add support for the metrics aggregations so they could be used easily like the current aggs:

`Product.search "pear", aggs: {average_price: {avg: :price}}` 

or with expanded options:

`Product.search "pear", aggs: {average_price: {avg: {field: :price}}`

Anyways, for the test would you like me to make one for setting aggs via body_options? Would that still go in the aggs_test?

ankane:
Yeah, a `body_options` test for aggs would be good. I think aggs_test is a fine place.

ankane:
Cleaning up PRs

Issue #840, opened 2017-02-08
Issue with randomization on version 2


RCReddy:
I am using elastic search from long time and I am happy to use it. 
I have upgraded the gem and elasticsearch to the latest version.
Earlier I am using elasticsearch 1.x and searchkick  0.8.7.
Random score is working fine in the older version but it is not working on latest version.
My query
query = Model.search("*",
      page: params[:page],
      per_page: params[:per_page] || PER_PAGE,
      operator: "or",
      fields:['title^20', 'tags'],
      where: filters,
      order: order_criteria,
      query: {function_score:{ random_score: { seed: DateTime.now.strftime("%Y%m%d").to_i } }},
      load: false,
      execute: false)
I got the following error in latest code.
 ArgumentError, "unknown keywords: query

ankane:
Hey @RCReddy, the `query` option was removed in Searchkick 2. Here's the preferred way to modify the query body: https://github.com/ankane/searchkick#advanced-search

PR #841, opened 2017-02-08
Issue with randomization on version 2 #840

RCReddy:
Please pull the code if would like it.


ankane:
Hey @RCReddy, thanks for the PR, but this was purposefully removed.

Issue #717, opened 2016-08-25
Searchkick 2.0 Plan


ankane:
`2.0_take3` branch

### Breaking

- [x] Remove `_all` field if searchable is specified
- [x] use keyword arguments
- [x] do not return array for `load: false`
- [x] add return value for `multi_search`

Clean up

- [x] clean up "private" methods in `Searchkick` module
- [x] remove `searchkick_create_index`, `clean_indices`, `searchkick_import`, `searchkick_debug`

Indexing improvements

- [x] allow `reindex` to be called on relation
- [x] rename `partial_reindex` to `reindex(:method_name)`

Remove old options / methods

- [x] replace `select` with `select_v2`
- [x] replace `merge_mappings` option with `default_mappings`
- [x] remove `query` and `json` search options (use `body` instead)
- [x] remove `include` option (use `includes` instead)
- [x] removed `personalize` option (use `boost_where` instead)
- [x] remove `partial` option (use `operator` instead)
- [x] remove legacy autocomplete option
- [x] remove facets (as we now have aggs)
- [x] drop support for ES 1
- [x] remove old `ReindexJob`

---

### Maybe

- [ ] Require fields to marked as `searchable` (or provided in the query)
- [ ] rename `searchkick_index` to `search_index`
- [ ] lazy searching by default (unsure if we want this)
- [ ] remove `searchkick_index_options` (replace with `mappings` proc)
- [ ] default to `misspellings: {below: 5}`
- [ ] removed latitude and longitude swap if passing coodinates as an array

### Non-breaking (can come after release)

- [x] add `debug` option to `search` method
- [x] add support for partial reindexes
- [x] add retry to class `reindex` method
- [ ] new `where` syntax with underscores - `_gt`, `_lt`, etc
- [ ] add `ReindexBatchJob` to support bulk reindex
- [ ] warn if you try to index an `id` field in `search_data`
- [ ] warn if `body` and conflicting options, like `where`

Comments welcome!

cireficc:
All of your proposed changes look good! I have a suggestion, but I don't know if it's warranted.

When reindexing associations, e.g. with a callback:

```ruby
after_commit :reindex associations

def reindex_associations
    users.each { |u| u.reindex }
    whatever.each { |w| w.reindex }
end
```

the above gets repetitive with the `each` call. Would it be possible/make sense to add the `reindex` method to the collection? So instead it would be:

```ruby
after_commit :reindex associations

def reindex_associations
    users.reindex
[...]

PR #851, opened 2017-02-16
Add removed options to changelog

rreinhardt9:
While adding I ran into the syntax change from `include` to `includes`. I added that option and some others that I noticed changed in #717 to the "breaking changes" section in the changelog and the readme to clarify for others that are working through the upgrade.

Can't wait to try out the goodies in 2.0...  especially queuing callback for batch updates!


ankane:
Thanks @rreinhardt9, great idea! `merge_mappings` was not changed, so I removed that.

PR #852, opened 2017-02-17
use custom query options like: analyzers, minimum_should_match

technopreneurG:
Feature: use custom set of analyzers instead of the default
Need: 
I could define custom analyzers by  
`
searchkick merge_mappings:true, 
  settings: {
    analyzer: {
      my_new_analyzer: {
        ...
      }
    }
  }  
`
But, I could not find a way to use it easily other than using body/body_options with writing whole query. With use_analyzers we can do: 
Product.search "abc def ghi", fields: [{'field1.analyzed^10': :custom}, {'field2.analyzed^5': :custom}], use_analyzers: ['my_new_analyzer', 'searchkick_word_search','searchkick_search','searchkick_search2']

Hope this is useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,10 @@ Breaking changes
 - Removed support for Elasticsearch 1 as it reaches [end of life](https://www.elastic.co/support/eol)
 - Removed facets, legacy options, and legacy methods
 - Invalid options now throw an `ArgumentError`
+- The `query` and `json` options have been removed in favor of `body`
+- The `include` option has been removed in favor of `includes`
+- The `personalize` option has been removed in favor of `boost_where`
+- The `partial` option has been removed in favor of `operator`
 - Renamed `select_v2` to `select` (legacy `select` no longer available)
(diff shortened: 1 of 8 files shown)


technopreneurG:
changes:
1) more options can be passed along with the analyzers, eg:
`
Product.search "abc def ghi", fields: [{'field1.analyzed^10': :custom}, {'field2.analyzed^5': :custom}], 
custom_query_options: [{analyzer: 'my_new_analyzer', minimum_should_match: '50%'}, {analyzer: 'searchkick_word_search', minimum_should_match: '50%'}, {analyzer: 'searchkick_search2', minimum_should_match: '50%'}]
`
2) Added tests

ankane:
Hey @technopreneurG, thanks for the PR!

I think it's better to use the `body` and `body_options` for something like this.

technopreneurG:
Hi, I started with body_options, then thought its good to provide easier ways to do the same for others also :) 

Let me know if you have any thoughts on further simplifying the syntax OR else please feel free to close without merging the PR!

PR #859, opened 2017-02-28
Adding a HashWrapper for results to prevent Hashie log warnings

mikelkew:
Recent updates to the `hashie` gem have resulted in a flood of log warnings when calling the `sort` method on results.

This was fixed recently in the `elasticsearch-rails` gem as per [this issue](https://github.com/elastic/elasticsearch-rails/issues/666).

Credit goes to @karmi whose implementation I've re-purposed here.

Sorry, I haven't written any tests for it.


ankane:
Hey @mikelkew, thanks for the PR 👍  Can you remove the `Results` namespace so it's `Searchkick::HashWrapper`?

mikelkew:
Sorry @ankane, had a crazy week and only just got to this. That's now been done, and some initial user testing on my end appears to be working fine.

ankane:
Looks great. After thinking a bit more, I'm a bit torn on if this is the right thing to do (even though other gems have taken this route). The warning was added for a reason, so it may be best to let developers silence as desired. Curious to hear your thoughts.

mikelkew:
Well, in this case it's Searchkick using and overriding the `sort` method. Realistically, to most users, the use of Hashie is somewhat transparent to them - their main concern is that it works and gives them the results in the format they expect.

If you are still hesitant though, perhaps it can be silenced by default in a new config option, but overridden if desired..?

karmi:
Just wanted to note I do agree with @mikelkew here -- I don't see a realistic situation where people would like to use the [Ruby-native `sort` method](https://ruby-doc.org/core-2.3.0/Enumerable.html#method-i-sort) for anything wrapper in a Hashie::Mash instance, those objects are usually "read-only". Of course, there _could_ be a case like that, but even in that case, the object should be wrapped in a custom class, not in a Hashie::Mash directly.

mikelkew:
Hi @ankane, just checking in to see if you've had a chance to give this any further consideration..?

ankane:
Thanks @mikelkew 👍

PR #727, opened 2016-09-09
Add support for cross-fields - fixes #707

sgerov:

PR #871, opened 2017-03-10
Added support for multi_match query with cross_fields type

technopreneurG:
Added support for multi_match query with cross_fields type.
example usage
`
Product.search "abc def ghi", fields: ['field1^20', 'field2^10'], cross_fields: true
`

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -17,7 +17,8 @@ module Searchkick
         :boost_by, :boost_by_distance, :boost_where, :conversions, :debug, :emoji, :exclude, :execute, :explain,
         :fields, :highlight, :includes, :index_name, :indices_boost, :limit, :load,
         :match, :misspellings, :offset, :operator, :order, :padding, :page, :per_page, :profile,
-        :request_params, :routing, :select, :similar, :smart_aggs, :suggest, :track, :type, :where]
+        :request_params, :routing, :select, :similar, :smart_aggs, :suggest, :track, :type, :where,
+        :cross_fields]
       raise ArgumentError, "unknown keywords: #{unknown_keywords.join(", ")}" if unknown_keywords.any?
 
       term = term.to_s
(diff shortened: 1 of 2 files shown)


ankane:
Hey @technopreneurG, thanks for the PR 👍  I have a good amount of feedback on the implementation, but I'm currently torn on the value of the feature after finding out about the `misspellings` limitation. My recommendation would be for devs to index search data into a single field if they want to search across fields.

jprosevear:
Indexing into a single field prevents per field boosting:
https://www.elastic.co/guide/en/elasticsearch/guide/current/_cross_fields_queries.html#_per_field_boosting

What is the misspellings limitation in this case?

@ankane do you have any specific architecture in mind for implementing this?

technopreneurG:
Hi @ankane , 

As mentioned in #727 
Quoting from https://www.elastic.co/guide/en/elasticsearch/guide/2.x/fuzzy-match-query.html  
>Tip
>Fuzziness works only with the basic match and multi_match queries. It doesn’t work with phrase matching, common terms, or cross_fields matches.

This PR should help with these use-cases:
1) Entity Search (as described in https://www.elastic.co/guide/en/elasticsearch/guide/current/_cross_fields_entity_search.html)
2) Per field boosting (as described in  https://www.elastic.co/guide/en/elasticsearch/guide/current/_cross_fields_queries.html#_per_field_boosting)

Do let me know if there are any changes in the implementation you would like to see.

For handling misspellings, I would assume spellings would have been already auto-corrected by using  https://github.com/ankane/searchkick#suggestions so that entities are correctly identified from the search query. Do let me know your suggestions. I can work on another PR for misspellings.

technopreneurG:
Hi @ankane,  Do you have any suggestions/or a different approach for the same?

ankane:
Hey @technopreneurG, thanks again for the PR. I don't see a big benefit for adding cross_fields right now. For the two use cases above, I recommend writing a custom query.

philipgiuliani:
@ankane Can you provide an example for a multi_match custom query with searchkick? Or should I just use the JSON api?

lucasmpereira:
@ankane I too, will be very grateful if you can provide an example of custom query to suport crossfield search?

marwenB:
@ankane we will be very grateful if you can provide an example of custom query to suport crossfield search.
Thank you

ankane:
Check out the [Advanced Search](https://github.com/ankane/searchkick#advanced-search) section for how to use custom queries. I don't have an example for cross fields, so please check out the Elasticsearch docs and share.

PR #876, opened 2017-03-21
Support cardinality metric aggregator

jprosevear:
Support cardinality metric aggregator - I suspect a next incremental step could be support for more metric aggregators by just maintaining a list of them (ie [:cardinality, :value_count...]), I think they are pretty much all the same in terms of query structure.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -643,6 +643,12 @@ module Searchkick
               interval: interval
             }
           }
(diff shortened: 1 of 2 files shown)


ankane:
Hey @jprosevear, thanks for the PR 💯 

Agreed we should find a more general way to address different agg types. I think we should try to find a way where we don't have to change Searchkick when new types are added to ES.

PR #881, opened 2017-03-29
Add `from` option to query

khusnetdinov:
Issue https://github.com/ankane/searchkick/issues/882


ankane:
Hey @khusnetdinov, thanks for the PR 👍 

You can use the limit and offset options to accomplish this already.

Issue #885, opened 2017-04-01
How to use wordnet on bonsai and heroku


jeremylynch:
The documentation discusses downloading wordnet and moving `wn_s.pl` to the `/var/lib`. 

However, how can this be achieved on `heroku` using `bonsai elasticsearch`? 

The only resource I have found discussing this is [this one](https://docs.bonsai.io/docs/using-wordnet-with-bonsai). 

The `wordnet_path` is set at `"/var/lib/wn_s.pl"` by default. Is it possible to change this to a directory within rails? ie within `public` or `tmp`?

rst0git:
Hi jeremylynch,

The wordnet_path seem to be hard-coded [here](https://github.com/ankane/searchkick/blob/f22aa73d81b2e6d978dbc1cb6526ce82af4df709/lib/searchkick.rb#L44)

You would be able to specify this path if the line
```
self.wordnet_path = "/var/lib/wn_s.pl"
```
is changed to something like:
```
elf.wordnet_path = ENV['WORDNET_PATH'] ? ENV['WORDNET_PATH'] : "/var/lib/wn_s.pl"
```

ankane:
Thanks for helping out @rst0git 👍

PR #889, opened 2017-04-08
Allow the user to specify the wordnet path

rst0git:
Using environment variable, allow the user to specify the wordnet path.
Issue [#885](https://github.com/ankane/searchkick/issues/885)


ankane:
Hey @rst0git, thanks for the PR! 👍  You can currently accomplish this with: 

```ruby
Searchkick.wordnet_path = 'some_path'
```

So I'd prefer not to add an extra `ENV` var.

PR #891, opened 2017-04-10
Match test mode when using parallel_spec

razum2um:
Hi, we're following these instructions to get parallel_spec + searchkick working: https://github.com/grosser/parallel_tests/wiki#with-searchkick----by-emaxi

This corrects behaviour across all running processes

### DIFF
--- a/lib/searchkick/index_options.rb
+++ b/lib/searchkick/index_options.rb
@@ -148,7 +148,7 @@ module Searchkick
           }
         }
 
-        if Searchkick.env == "test"
+        if Searchkick.env =~ /test/
           settings[:number_of_shards] = 1
           settings[:number_of_replicas] = 0
         end


ankane:
Hey @razum2um, thanks for the PR 👍 

Instead of changing the `env`, I think it'd be better to add an `index_suffix` option, so you could do:

```ruby
Searchkick.index_suffix = ENV["TEST_ENV_NUMBER"]
```

ankane:
Hey @razum2um, I added an `index_suffix` option and tested it with parallel tests. Thanks again for the PR.

razum2um:
@ankane cool, thanks for push, didn't get time to submit PR like you wished. When could you release it?

ankane:
A new version was published earlier today.

razum2um:
Thanks for a lot for the gem and such responsiveness! ❤️

PR #901, opened 2017-04-20
Use an after create hook instead of manual reindex

tilsammans:
I always use FactoryGirl's `after create` hook to reindex. Since typically I only create only a handful models needed for each test, this is very quick and I don't have to remember anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1769,18 +1769,25 @@ end
 
 ### Factory Girl
 
-Use an after `create` hook for each indexed model:
+Use a trait and an after `create` hook for each indexed model:
 
 ```ruby
 FactoryGirl.define do
   factory :product do
     # ...
-    
-    after(:create) do |product, evaluator|
-      Product.reindex
+
+    # Note: This should be the last trait in the list so `reindex` is called
+    # after all the other callbacks complete.
+    trait :reindex do
+      after(:create) do |product, _evaluator|
+        product.reindex(refresh: true)
+      end
     end
   end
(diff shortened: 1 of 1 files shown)


ankane:
Hey @tilsammans, thanks for the PR 👍 

Is there a good way to do this conditionally? It won't be great for performance if you're reindexing records when you're not testing search.

elskwid:
@tilsammans & @ankane,

I've had some success using FactoryGirl a `trait` for this so that it's only done when needed/requested.

```ruby
# factory
FactoryGirl.define do
  factory :product do
    # ...

    # Note: This should be the last trait in the list so `reindex` is called
    # after all the other callbacks complete.
    trait :reindex do
      after(:create) do |product, _evaluator|
        product.reindex(refresh: true)
      end
    end
  end
end

# use it
FactoryGirl.create(:product, :some_trait, :reindex, some_attribute: "foo")
```

ankane:
Hey @elskwid, that sounds like a great approach.

tilsammans:
That seems like a good approach! I'll amend the PR and rebase.

I've had really no performance issues btw, but this seems cleaner.

tilsammans:
@ankane all done! thanks @elskwid

ankane:
Great, thank you both 💯

elskwid:
👍  Glad I could help. Thanks for the PR @tilsammans and thanks for the project @ankane. Take care!

Issue #907, opened 2017-04-29
Test fail on Elasticsearch 5.3.1


titanmaru:
As in ES version 5.3.1 [release notes](https://www.elastic.co/guide/en/elasticsearch/reference/current/release-notes-5.3.1.html), they rejected any bulk index with empty IDs which cause one test to fail.

ankane:
Fixed by #908 🎉

PR #908, opened 2017-04-29
fix bulk index without IDs ankane/searchkick#907

titanmaru:
This will create products before trying to index in ES which prevent ES to complain that products do not have valid ID.


ankane:
Great, thanks @titanmaru 👍

Issue #799, opened 2017-01-05
Question: Customize queue name


jkras:
Hey @ankane, firstly, thanks for searchkick. It has made life incredible! 

Any idea how I could overwrite the queue name for reindex_async? 

I did a little sniffing and found that the queue is labelled below, though am still reasonably knew to rails and would like to know the most appropriate way to customize this given I'm using your gem, rather than a job I've written myself.   

`class ReindexV2Job < ActiveJob::Base
    queue_as :searchkick`

Thanks in advance

Jesse

ankane:
Searchkick doesn't provide a way, but you can create an initializer with:

```ruby
Searchkick::ReindexV2Job.queue_as :other_name
```

If someone wants to put together a PR for a `Searchkick.queue_name` method, I'll be happy to accept.

jessekrasnostein:
Thank you. I'd love to give the PR a crack, any tips or pointers on where to start. It would be my first open source contribution.

ankane:
Cool, you should be able to add it in a similar fashion to [these methods](https://github.com/ankane/searchkick/blob/64e350049947d1965b97f928af35ae89df6d512e/lib/searchkick.rb#L33). To make sure nothing broke, run [the tests](https://github.com/ankane/searchkick#contributing). For this feature, I don't think we need a new test. When you're done, create a PR and I'll take a look.

For general PR advice, check out the [GitHub docs](https://guides.github.com/activities/contributing-to-open-source/#contributing).

ankane:
Added in #909

PR #909, opened 2017-04-29
Add customizable queue name

titanmaru:
As requested [here](https://github.com/ankane/searchkick/issues/799).  I also added how to change queue name in readme as well :)  Hope everyone will be happy.


ankane:
Thanks again @titanmaru 💯

PR #933, opened 2017-06-13
Add support for conversions boost term

sarchak:
**Why?**

Currently, when boosting on conversions we use the term for boosting. This PR will add support to have separate conversion boost term.

One way to improve search results is if we can have separate terms for matching and boosting.

Example :
   *  organic banana:
      * Entity : banana
      * Attribute: organic

We need a way to use the existing conversions for boosting (using popularity for boosting) and be more flexible with matching. In our case match would be with "banana" and conversions boosting will be with "organic banana"

cc @ankane @meetrajesh

### DIFF
--- a/README.md
+++ b/README.md
@@ -1587,9 +1587,10 @@ end
 and during query time:
 
 ```ruby
-Product.search("banana") # boost by both fields (default)
+Product.search("banana") # boost by both fields (default) and term (banana) will be used for both matching and boosting.
 Product.search("banana", conversions: "total_conversions") # only boost by total_conversions
 Product.search("banana", conversions: false) # no conversion boosting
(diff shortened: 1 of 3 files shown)


ankane:
Hey @sarchak, thanks for the PR, including tests 💯  Can we call it `conversions_term` since it's shorter?

sarchak:
@ankane by default ```term``` is used for dual purpose (match and boost) I, wanted to be very specific and tell that this would only be used for boosting. I am open to ```conversions_term``` but I think having the boost in the term will make it a bit clear for the user.

ankane:
Let's go with `conversions_term`, as everything around conversions deals with boosting, not matching.

sarchak:
Done!

ankane:
Looks great, thanks! 👍

PR #940, opened 2017-06-16
maintain boost oreder in where query active record

chitrayadav91:
@ankane, In my project I was using searchkick version 1.3.1. As per the requirement we need to boost title more then description so I used the boost by title in query like - 

`FeedItem.search(query, fields: ["title^3", "description^1"]`

But I was not getting the right order after boosting so I debug the issue and found a small issue in gem that It does not maintain the Active Record Order after boosting.

I just fixed the issue on branch 'boost_active_record_ordering'. I created this branch from tag v1.3.1.

### DIFF
--- a/lib/searchkick/results.rb
+++ b/lib/searchkick/results.rb
@@ -189,8 +189,7 @@ module Searchkick
 
       if records.respond_to?(:primary_key) && records.primary_key
         # ActiveRecord
-        records = records.where(records.primary_key => ids)
-        ids.collect {|id| records.detect {|x| x.id == id.to_i}}.compact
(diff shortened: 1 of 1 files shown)


ankane:
Hey @chitrayadav91, thanks for the PR 👍 

The order is maintained in Ruby here: https://github.com/ankane/searchkick/blob/599c1d5fac678f85dae18fd7e82fb4b0965dae11/lib/searchkick/results.rb#L34

chitrayadav91:
Thanks @ankane. Now I am using .results instead of .records. Now order is maintaining in older version 1.3.1. Thanks for your quick support. :)

ankane:
Hey @chitrayadav91, the `records` method will likely be removed in Searchkick 3, so going to close this PR.

PR #946, opened 2017-06-29
Fix non-numeric async reindex

danielwestendorf:
Calling `ModelScope.reindex(async: true)` fails when non-numerical primary keys are in play, such as UUID’s.

Unless the primary key has a database column type of integer, float, or
decimal, index in batches by plucking the primary key column and
sending it to the job as record_ids.

### DIFF
--- a/lib/searchkick/index.rb
+++ b/lib/searchkick/index.rb
@@ -420,14 +420,23 @@ module Searchkick
       if scope.respond_to?(:primary_key)
         # TODO expire Redis key
         primary_key = scope.primary_key
-        starting_id = scope.minimum(primary_key) || 0
-        max_id = scope.maximum(primary_key) || 0
-        batches_count = ((max_id - starting_id + 1) / batch_size.to_f).ceil
 
-        batches_count.times do |i|
-          batch_id = i + 1
-          min_id = starting_id + (i * batch_size)
-          bulk_reindex_job scope, batch_id, min_id: min_id, max_id: min_id + batch_size - 1
+
+        if %i(integer float decimal).include?(scope.column_for_attribute(primary_key).sql_type_metadata)
+          starting_id = scope.minimum(primary_key) || 0
+          max_id = scope.maximum(primary_key) || 0
+          batches_count = ((max_id - starting_id + 1) / batch_size.to_f).ceil
+
+          batches_count.times do |i|
+            batch_id = i + 1
+            min_id = starting_id + (i * batch_size)
+            bulk_reindex_job scope, batch_id, min_id: min_id, max_id: min_id + batch_size - 1
+          end
+        else
+          batch_id = 1
+          scope.in_batches(of: batch_size).each do |batch|
+            bulk_reindex_job scope, batch_id, record_ids: batch.pluck(primary_key)
+          end
         end
       else
         batch_id = 1
--- a/lib/searchkick/index.rb
+++ b/lib/searchkick/index.rb
@@ -422,7 +422,7 @@ module Searchkick
         primary_key = scope.primary_key
 
 
-        if %i(integer float decimal).include?(scope.column_for_attribute(primary_key).sql_type_metadata)
+        if scope.minimum(primary_key).is_a?(Numeric)
           starting_id = scope.minimum(primary_key) || 0
           max_id = scope.maximum(primary_key) || 0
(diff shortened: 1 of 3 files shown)


danielwestendorf:
Fixes #947

test/test_helper.rb
```diff
@@ -477,6 +504,13 @@ class Animal
     # wordnet: true
 end
 
+class Sku
+  searchkick \
```
ankane: I don't think any of the options are necessary.

lib/searchkick/index.rb
```diff
@@ -420,14 +420,23 @@ def full_reindex_async(scope)
...
-        max_id = scope.maximum(primary_key) || 0
-        batches_count = ((max_id - starting_id + 1) / batch_size.to_f).ceil
-
-        batches_count.times do |i|
-          batch_id = i + 1
-          min_id = starting_id + (i * batch_size)
-          bulk_reindex_job scope, batch_id, min_id: min_id, max_id: min_id + batch_size - 1
+
+
+        if %i(integer float decimal).include?(scope.column_for_attribute(primary_key).sql_type_metadata)
```
ankane: Two things:

1. How consistent is `sql_type_metadata` across different database adapters?
2. Not sure the list is complete. Does it cover bigints?

lib/searchkick/index.rb
```diff
@@ -420,14 +420,23 @@ def full_reindex_async(scope)
...
+
+          batches_count.times do |i|
+            batch_id = i + 1
+            min_id = starting_id + (i * batch_size)
+            bulk_reindex_job scope, batch_id, min_id: min_id, max_id: min_id + batch_size - 1
+          end
+        else
+          batch_id = 1
+          scope.in_batches(of: batch_size).each do |batch|
+            bulk_reindex_job scope, batch_id, record_ids: batch.pluck(primary_key)
```
ankane: Any reason `batch_id` is always 1?

lib/searchkick/index.rb
danielwestendorf: 1. I was concerned about this as well, I'm not sure how consistent it is. It seems to be a generic API that can be over ridden by specific adapters if needed.
2. I don't think so. I thought I'd submit the PR before putting the time into investigating this. I'll do some research and get it updated.

danielwestendorf: The inclusive nature of this for "typical" use is probably 👎 . It's not future proof. I'll look into alternative methods for implementing this. There has to be a way to tell if a column is a Numeric data type, right? 🤷‍♂️

lib/searchkick/index.rb
danielwestendorf: I think another problem here may be that `in_batches` is relatively (no pun intended) new to AR, introduced in rails5. I assume rails4.x support is needed here.

lib/searchkick/index.rb
```diff
@@ -420,14 +420,24 @@ def full_reindex_async(scope)
...
+          max_id = scope.maximum(primary_key) || 0
+          batches_count = ((max_id - starting_id + 1) / batch_size.to_f).ceil
+
+          batches_count.times do |i|
+            batch_id = i + 1
+            min_id = starting_id + (i * batch_size)
+            bulk_reindex_job scope, batch_id, min_id: min_id, max_id: min_id + batch_size - 1
+          end
+        else
+          scope.in_batches(of: batch_size).each_with_index do |batch, i|
```
danielwestendorf: `#in_batches` was introduced in rails5. I think this needs 4.x support, so probably `#find_in_batches` instead and collecting the PK's in code... 😢

ankane: Yeah, `find_in_batches` would be better.

lib/searchkick/index.rb
ankane: One idea would be to call `scope.minimum(primary_key)` and check if it's numeric.

danielwestendorf:
@ankane I've resolved all the concerns we had on this PR. Thanks for the feedback.

ankane:
Awesome, thanks @danielwestendorf, looks great! 💯  Made a small change here for performance https://github.com/ankane/searchkick/commit/8b6d067e2fcbd5b0dc1c6ec3d461c78e4a0711c7

PR #950, opened 2017-07-05
Pre- and post-indexing callbacks

will-r:
I don't know if this is going to be of interest, but I've found it useful in a couple of cases where pubsub notifications should be sent out only after indexing is complete.


ankane:
Hey @will-r, thanks for the PR 👍  I like the idea. To keep it consistent, I think we should use `reindex` instead of `index`. Also, instead of having `before_index` and `after_index` options, users should be able to do:

```ruby
class User
  after_reindex :publish_something

  def publish_something
  end
end
```

I think you'll need to add `run_callbacks :reindex` around the `reindex` method to accomplish this.

Lastly, I don't think the tests actually cover this.

ankane:
Cleaning up stale PRs

Issue #701, opened 2016-07-12
Feature Request: Eager loading for multiple models


reidab:
I'm searching across multiple multiple models and would like to eager-load associations. The current implementation fails when it tries to send `includes` to the wrong model class.

The current implementation of Searchkick's `include` option takes an array of associations to eager load (`Searchkick.search "milk", include: [:brand, :stores]`). What would you say to a PR to optionally accept a hash, like so?

```
Searchkick.search "milk", index_name: [Product, Category], include: {Product => [:brand, :stores]}
```

ankane:
Hey @reidab, I'm open to the idea. I think we'd want another option to accomplish this to keep backward compatibility and not confuse people.

tracyloisel:
@ankane I can help for this feature. I see a way to succeed :
for each model in index_name, build a search query with option `execute: false`, then call Searchkick.multi_search

After reading results.rb#L186 I suggest a new option could be "includes_at", like that

`Search.search "milk", index_name: [Product, Category], includes_at: {product: [:brand,:stores], }

all_search_query = index_name.collect do |domain|
domain.search "milk", include: options[:includes_at].fetch(domain.to_sym), execute: false
end`

then like in https://github.com/ankane/searchkick#multi-search :
`Searchkick.multi_search(all_search_query)`

ankane:
@tracyloisel Thanks for offering to help. You should be able to make the change at results.rb#L186 without the need for `multi_search`.

oyeanuj:
@reidab @tracyloisel Curious, if a PR for this still on your radars?

tracyloisel:
Yes @oyeanuj . Do you need that feature soon ?

oyeanuj:
@tracyloisel yes, that would be amazing!

twanies:
Any updates yet on the status of this pr?

PR #967, opened 2017-07-28
added per-model includes support

Xtinct:
Implements #701

### DIFF
--- a/test/sql_test.rb
+++ b/test/sql_test.rb
@@ -1,5 +1,4 @@
 require_relative "test_helper"
-
 class SqlTest < Minitest::Test
   def test_operator
     store_names ["Honey"]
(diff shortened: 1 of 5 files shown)


ankane:
Hey @Xtinct, thanks for the PR 👍  As is, this breaks backwards compatibility as `includes` currently accepts a hash for nested associations. I think we'll need to add a new option.

Xtinct:
Hey @ankane, thanks for reviewing. Added option called `includes_per` to handle separate includes.

test/test_helper.rb
```diff
@@ -84,6 +84,14 @@ class Product
     field :alt_description
   end
 
+  class Discount
```
ankane: Can we get away without adding a new model?

test/sql_test.rb
```diff
@@ -1,5 +1,4 @@
 require_relative "test_helper"
-
```
ankane: Please don't change whitespace.

Xtinct:
Hey @ankane, updated according to your comments.

ankane:
Looks great, thanks @Xtinct 💯  Wasn't super happy with `includes_per_model` (despite suggesting it), so I changed the name to `model_includes`. Will cut a new release with this very soon.

PR #968, opened 2017-07-28
Adds an optional 'missing' param to boost_by

jordanderson:
Simply adds `missing` key and value to `field_value_factor`, which helps avoid `Searchkick::InvalidQueryError` on unmapped fields. 

Let me know what you think!

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -872,7 +872,7 @@ module Searchkick
           }
         }
 
-        if value[:missing].present?
+        if value[:missing].present? && !below50?
           script_score[:field_value_factor].merge!({missing: value[:missing].to_f})
         end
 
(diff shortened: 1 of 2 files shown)


jordanderson:
I thought that missing param was supported in field_value_factor since elasticsearch 2.0, but perhaps not... 

Is this change useful to anyone? Our use case is that we have some dynamically-generated fields and occasionally we boost the value in one before it has been created. This is a nice way to continue to do this and eliminate a check to make sure the field exists prior to every query or a rescue block where we delete that boost_by if it raised a `Searchkick::InvalidQueryError` exception and then requery.

Feel free to close this if it's not generally helpful.

ankane:
Hey @jordanderson, thanks for the PR 👍  Makes sense to me. Let's throw an `ArgumentError` for ES < 5 to it's clear the option doesn't work.

ankane:
Thanks @jordanderson. All merged! I made a few changes here: https://github.com/ankane/searchkick/commit/3b628b0d11271340b65d519ce8424fce092951db

jordanderson:
Thanks!

PR #981, opened 2017-08-31
Prevents ES from sending gzip'd responses.

theschoolmaster:
This is a simple fix for AWS ES new behavior to send gzipped responses when Accept-Encoding is not set.  

An alternative approach would to enable FaradayMiddleware::Gzip.  This would automatically set Accept-Encoding to “gzip,deflate”.

See https://github.com/elastic/elasticsearch-ruby/issues/457 for slightly more info.

### DIFF
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -53,7 +53,7 @@ module Searchkick
 
       Elasticsearch::Client.new({
         url: ENV["ELASTICSEARCH_URL"],
-        transport_options: {request: {timeout: timeout}, headers: {content_type: "application/json"}}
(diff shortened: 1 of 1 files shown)


ankane:
Hey @theschoolmaster, thanks for the PR and the context 👍  I'd like to wait a bit to see how things shake out. As @dinusX mentioned, it looks like the signer middleware is requesting `gzip`.

ankane:
Looks like v0.1.9 fixes this 🎉  https://github.com/winebarrel/faraday_middleware-aws-signers-v4/pull/14

PR #994, opened 2017-09-20
Use normal filters instead of post filters to handle smart_aggs queries

elguapo1611:


Schwad:
Is this matched with a specific issue?

ankane:
@elguapo1611 and I discussed offline and decided a more general approach would be find which filters could be applied pre and post. Will give this a shot when I have chance. Def a performance hit with the existing implementation.

PR #996, opened 2017-09-20
Allow operator to be a symbol or string

elguapo1611:


ankane:
Awesome, thanks @elguapo1611 👍

PR #1000, opened 2017-09-27
Added support for custom field_value_factor modifier.

### LINKED ISSUE #498: field_value_factor modifier
Hi all, 
Is there any way to specify the field_value_factor modifier in the boost_by parameter?

Example:
boost_by: {orders_count: {factor: 10, modifier: sqrt}}

Currently, the searchkick seems to use the ln2p as a default modifier.

Any help will be appreciated.

itay-grudev:
The implementation maintains backwards compatability with the `log` parameter,
though it should be depricated.

This fixes #498.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -873,13 +873,11 @@ module Searchkick
 
     def boost_filters(boost_by, options = {})
       boost_by.map do |field, value|
-        log = value.key?(:log) ? value[:log] : options[:log]
-        value[:factor] ||= 1
         script_score = {
           field_value_factor: {
             field: field,
-            factor: value[:factor].to_f,
-            modifier: log ? "ln2p" : nil
+            modifier: (value[:modifier] or ( (value[:log] or options[:log]) ? 'ln2p' : nil ) or 'none'),
+            factor: (value[:factor] ? value[:factor].to_f : 1),
           }
         }
 
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -873,10 +873,11 @@ module Searchkick
 
     def boost_filters(boost_by, options = {})
       boost_by.map do |field, value|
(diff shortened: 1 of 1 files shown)


lib/searchkick/query.rb
```diff
@@ -873,13 +873,11 @@ def custom_filter(field, value, factor)
...
       boost_by.map do |field, value|
-        log = value.key?(:log) ? value[:log] : options[:log]
-        value[:factor] ||= 1
         script_score = {
           field_value_factor: {
             field: field,
-            factor: value[:factor].to_f,
-            modifier: log ? "ln2p" : nil
+            modifier: (value[:modifier] or ( (value[:log] or options[:log]) ? 'ln2p' : nil ) or 'none'),
+            factor: (value[:factor] ? value[:factor].to_f : 1),
```
itay-grudev: The change for the factor field is purely aesthetic.

itay-grudev: Also since the `ln2p` was uses when either `value[:log]` or `options[:log]` are `true` a simple `or` statement was a better implementation. Additionally when there is no modifier set this will correctly default to `'none'` as per the ES documentation.

itay-grudev: And finally with this modification added I recommend that you deprecate the `log` option.

ankane:
Hey @itay-grudev, thanks for the PR 👍  This is a great addition. A few comments:

1. Try to keep the changeset to a minimum (no unnecessary changes)
1. `or` instead of `key?` changes the logic for `log: false`
2. In Ruby, `or` is used as a flow control operator - use `||` for boolean operations

lib/searchkick/query.rb
itay-grudev: The factor refactoring uses less memory this way and is a 1-liner. Can it stay?

ankane:
Hey @itay-grudev, I added support on master.

PR #1002, opened 2017-09-27
Added support for the tie_breaker argument for dis_max queries

itay-grudev:
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-dis-max-query.html

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -19,7 +19,7 @@ module Searchkick
         :boost_by, :boost_by_distance, :boost_where, :conversions, :conversions_term, :debug, :emoji, :exclude, :execute, :explain,
         :fields, :highlight, :includes, :index_name, :indices_boost, :limit, :load,
         :match, :misspellings, :model_includes, :offset, :operator, :order, :padding, :page, :per_page, :profile,
-        :request_params, :routing, :select, :similar, :smart_aggs, :suggest, :track, :type, :where]
+        :request_params, :routing, :select, :similar, :smart_aggs, :suggest, :track, :type, :where, :tie_breaker]
       raise ArgumentError, "unknown keywords: #{unknown_keywords.join(", ")}" if unknown_keywords.any?
 
       term = term.to_s
@@ -372,7 +372,8 @@ module Searchkick
 
           payload = {
             dis_max: {
-              queries: queries
+              queries: queries,
+              tie_breaker: (options[:tie_breaker] or 0)
             }
           }
 


ankane:
Hey @itay-grudev, can you give more context on this?

itay-grudev:
> The tie breaker capability allows results that include the same term in multiple fields to be judged better than results that include this term in only the best of those multiple fields, without confusing this with the better case of two different terms in the multiple fields.The default tie_breaker is 0.0.
> – [Elastic Search Reference](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-dis-max-query.html)

It's a feature of DisMax queries. Check the link for a complete explanation.

ankane:
Searchkick's general approach is to provide higher-level/abstracted search concepts rather than specific Elasticsearch features. It feels like this may fall into the latter. Can you give a practical example where this is useful? It'll also need test case before merging.

itay-grudev:
I get that but sometimes you need to customise queries a bit. Unless I apply all this I get mostly crappy results based only on the object title:

Example: Property search for jQuery UI autocomplete:

```ruby
    @offices = Office.search(
      params[:term],
      fields: [
        { 'name^10' => :word_start },
        { 'postcode^9' => :word_start },
        { 'town^7' => :word_start },
        { 'street^5' => :word_start },
        { 'tags^3' => :word },
        { 'county^1' => :word },
        { 'description^1' => :word },
      ],
      boost_by: {
        capacity: {
          factor: 0.1,
          modifier: 'none',
          missing: 1,
          boost_mode: 'sum',
        },
      },
      tie_breaker: 1 / Math::E,
      match: :word_start,
    )
```

Note: This example is not realistic as I am not at liberty to share the scope of the app I am working on.

Now here is some description about why I need all this:

1. **`modifier`** and **`boost_mode`:** I need to affect the results score just with a relatively small number directly proportional the capacity of the property because statistically larger properties have more personnel and are more likely to be of relevance. I need the `modifier` to be linear aka `none` and the factor `0.1` based on the average capacity. I also need to add this to the existing score and not multiply with a logarithmic dependency as that changes the score too much. Ideally if two offices have exactly the same fields, but one of them is slightly larger it should come first with very slightly higher score.
2. **`tie_breaker`:** Assuming that I am looking for `London` and I have 3000 results with have 1 reference to `London` in their title or street they will get relatively equal scores. But a result that has `London` in multiple fields like `title` and `town` is much more relevant to me than one that just has a match in the title field. Thus I need to score those results more and that is what `tie_breaker` does. Based on some clever maths and statistics `1/e` (≈0.37) actually does a very good job.

itay-grudev:
@ankane I generally like how simple it is to build queries with Searchkick but I really need more control over the result scoring and that's why I've spammed you so much over the last 24 hours.

ankane:
Hey @itay-grudev, sorry for the delay. If you're still interested in this, please add a test or two.

itay-grudev:
I'm interested but I don't have the time at the moment. I'm really sorry.

ankane:
No worries, hope it's not too stressful. Feel free to reopen when you have the time.

PR #1014, opened 2017-10-23
Set ignore_above option based on ES version

matcouto:
- Set **ignore_above** option only for Elasticsearch versions above 2.2.0 and when it's not a field of **text** type.
- It seems that this option is not valid for **text** field type on Elasticsearch version greater than 2.2.0 according to the links below:
https://stackoverflow.com/questions/41068670/elasticsearch-5-0-2-ignore-above-gives-unsupported-parameter-error
https://www.elastic.co/guide/en/elasticsearch/reference/master/ignore-above.html

Thoughts?

### DIFF
--- a/lib/searchkick/index_options.rb
+++ b/lib/searchkick/index_options.rb
@@ -9,9 +9,8 @@ module Searchkick
         settings = options[:settings] || {}
         mappings = options[:mappings]
       else
-        below22 = Searchkick.server_below?("2.2.0")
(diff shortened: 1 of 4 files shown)


ankane:
Hey @matcouto, thanks for the PR 👍 The ignore_above option is only set for keyword mappings currently, so don't think this is needed. Also, for future PRs (for any open source projects), it's best to keep the change set as small as possible and submit multiple PRs if there are multiple changes you'd like to make.

PR #1017, opened 2017-10-24
Separate emoji parse method

tehPlayer:
Separating emoji parse method allows developer to overwrite it.

Scenario:
If developer will store in ES parsed emoji in other way e.g. `[[crown]]` or `:crown:` instead `crown`, then let him overwrite this method, to use new format.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -24,9 +24,7 @@ module Searchkick
 
       term = term.to_s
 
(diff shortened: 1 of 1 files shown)


ankane:
Hey @tehPlayer, thanks for the PR 👍 Internal methods should be expected to change, so a better approach would be to do any custom query transformations before passing it to Searchkick.

PR #1020, opened 2017-11-03
Add select: {exclude... } to readme

jordanderson:
A helpful option I discovered looking through the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,13 +130,6 @@ Select
 select: [:name]
 ```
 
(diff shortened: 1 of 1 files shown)


ankane:
Hey @jordanderson, thanks for the PR 👍  I think we could be more general and link to [this page](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-source-filtering.html).

ankane:
Awesome, thanks @jordanderson 👍

PR #1022, opened 2017-11-09
Handle sparse tables

outerim:
Some tables have large swaths of missing IDs so the default bulk reindexing strategy is extremely inefficient. In these cases falling back to find_in_batches, as if we didn't have numeric IDs in the first place is an appropriate compromise.

I wasn't able to find a test that covered this branch of Numeric batch handling so I didn't write a test 😱. If someone wants to point me in the right direction I'd be happy to mend my mistake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1667,6 +1667,15 @@ class Product < ActiveRecord::Base
 end
 ```
 
+Handle sparse tables
+
+```ruby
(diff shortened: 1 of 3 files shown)


outerim:
Meh, closing this because find_in_batches with various layers of includes doesn't seem to be able to feed queues fast enough... Will re-evaluate.

PR #1028, opened 2017-11-20
Enhance boost by distance with factor

pplant:
Added the possiblity to set a factor that is handed over as a weight field for the boost_by_distance function. This allows to mix different types of queries defining a factor. 

`Order.search 'coffee', boost_where: { total_price: {value: 100, factor: 70}, boost_by_distance: {
  field: :ordered_at,
  function: :gauss,
  origin: Time.now,
  scale: '1w',
  offset: '0',
  decay: 0.5,
  factor: 100
}`

In this case instead of the order's scores calculated by the gauss function being a numeric value ranging from 0-1 will be multiplied by 100 and therefore its weight will be higher than this of the total_price.


pplant:
Any news on this one?

ankane:
Thanks @pplant 👍  Happy to include the feature, but the test doesn't really test it.

pplant:
Hi, I will enhance the test to check the resulting scores.

ankane:
Beautiful, thanks @pplant 💯

Issue #843, opened 2017-02-09
Reindex of Collection gives Elasticsearch Transport error


Ravi1729:
Hello,

I have a model with name **Actor**.And the model looks like the following snippet

```
class Actor 
      include Mongoid::Document  
      field :name, type: String 
      field :birthName, as: :birth_name, type: String 
      field :date_of_birth, type: Date 
      field :height, type: Measurement 
      field :bio, type: String 
    searchkick
end
```

when I try to reindex my model using  `Actor.reindex`. I am getting the following error which shows the root cause as **Index already exists**.

```
irb(main):001:0> Actor.reindex
Elasticsearch::Transport::Transport::Errors::BadRequest: [400] {"error":{"root_cause":[{"type":"index_already_exists_exception","reason":"index [actors_development_20170209175219593/TKOPkMY1TsGNEYmWlVCKo
A] already exists","index_uuid":"TKOPkMY1TsGNEYmWlVCKoA","index":"actors_development_20170209175219593"}],"type":"index_already_exists_exception","reason":"index [actors_development_20170209175219593/TKO
PkMY1TsGNEYmWlVCKoA] already exists","index_uuid":"TKOPkMY1TsGNEYmWlVCKoA","index":"actors_development_20170209175219593"},"status":400}
        from /home/ravi/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/elasticsearch-transport-5.0.1/lib/elasticsearch/transport/transport/base.rb:201:in `__raise_transport_error'
        from /home/ravi/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/elasticsearch-transport-5.0.1/lib/elasticsearch/transport/transport/base.rb:318:in `perform_request'
        from /home/ravi/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/elasticsearch-transport-5.0.1/lib/elasticsearch/transport/transport/http/faraday.rb:20:in `perform_request'
        from /home/ravi/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/elasticsearch-transport-5.0.1/lib/elasticsearch/transport/client.rb:128:in `perform_request'
        from /home/ravi/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/elasticsearch-api-5.0.1/lib/elasticsearch/api/namespace/common.rb:21:in `perform_request'
[...]

PR #1034, opened 2017-12-01
Fix a race condition during reindex

kakra:
As outlined in https://github.com/ankane/searchkick/issues/843
reindexing may sometimes fail with 'index_already_exists_exception'.

This primitive patch fixes it by retrying the index creation up to 10
times following a fibonacci timeout curve, that is on failure it waits
in seconds: 1, 1, 2, 3, 5, 8, 13, 21, 34, 55 before giving up.
Reindexing is a slow process anyways, so we can wait around 3 minutes
(including time needed for index deletion) before it finally fails.

For 1.5.1, please pull from kakra hotfix-1.5.1/fix-exception-on-reindex.

Signed-off-by: Kai Krakow <kai@kaishome.de>

### DIFF
--- a/lib/searchkick/index.rb
+++ b/lib/searchkick/index.rb
@@ -226,11 +226,25 @@ module Searchkick
         raise Searchkick::Error, "No index to resume" unless index_name
         index = Searchkick::Index.new(index_name)
       else
-        clean_indices unless retain
-
         index_options = scope.searchkick_index_options
         index_options.deep_merge!(settings: {index: {refresh_interval: refresh_interval}}) if refresh_interval
-        index = create_index(index_options: index_options)
+
+        failed, timeout, next_timeout = 0, 0, 1
+        begin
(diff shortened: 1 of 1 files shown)


ankane:
Hey @kakra, thanks for the PR 👍 You really don't want to be creating multiple indexes per second (which I think is when this is triggered). I think this probably belongs in app-specific code.

kakra:
Yes, I really don't want to, but my code didn't try so... I have no idea where the race comes from, but putting in some waiting seems to fix it. Tho, I'm not sure because it's not always reproducible. I guess elasticsearch is maybe delaying some actions and then it starts to race with the app doing multiple stuff in _sequential_ order. I was really only running "rake searchkick:reindex:all" when the issue occurs, everything else is idle.

ankane:
Hey @kakra, thanks again for the PR. Going to close as I think app-specific code can handle this.

PR #1037, opened 2017-12-05
Add `histogram` aggregation

divineforest:
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-histogram-aggregation.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Better exception when trying to access results for failed multi-search query
 - Added support for `faraday_middleware-aws-sigv4`
 - Added `credentials` option to `aws_credentials`
+- Added `histogram` aggregation
 
 ## 2.4.0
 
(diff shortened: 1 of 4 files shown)


ankane:
Hey @divineforest, thanks for the PR, including tests 💯  Since Elasticsearch has a lot of different aggregation types, I'd prefer not to add them one by one, but instead have developers use `body_options`.

```ruby
Product.search "pear", body_options: {aggs: {prices: {histogram: {field: :price, interval: 10}}}
```

I'll update the readme to make this more clear.

divineforest:
@ankane but since `date_histogram` is already there a regular `histogram` seems like it is missing :(

ankane:
@divineforest Other aggregation types were added before this decision was made. Have to start somewhere.

divineforest:
@ankane ok, thanks for the explanation :)

PR #1041, opened 2017-12-12
Use Searchkick.queue_name for bulk indexing jobs

kennon:
ReindexV2Job uses Searchkick.queue_name for its ActiveJob queue, but the bulk indexing jobs are hardcoded to :searchkick (the default). This PR uses Searchkick.queue_name for the bulk jobs as well for consistency.


ankane:
Awesome, thanks @kennon 👍

Issue #893, opened 2017-04-11
Add db scopes to :async ReindexV2Job


esiegel:
It would be nice if the library didn't assume a record will be found using the default scope.  Specifically this line in `ReindexV2Job` causes an issue.

```ruby
model.find(id)
```

We are using the ruby library [paranoia](https://github.com/rubysherpas/paranoia), which soft destroys records such that the active record default scope won't include that record.  We would need to query the model like so:

```ruby
model.with_deleted.find(id)
```

Maybe we can reuse searchkick's `search_import` scope or maybe add another.  Or the job could be written in a way to allow overriding.

ankane:
Hey @esiegel, you can use the should_index? method to prevent a record from being indexed. https://github.com/ankane/searchkick#indexing

esiegel:
@ankane I actually **do** want to index the record, and the default implementation of `should_index` is fine as it indexes everything.  I need to be able to search ElasticSearch for these soft destroyed records.

The issue is that during `:async` reindexing the job cannot [find](https://github.com/ankane/searchkick/blob/master/lib/searchkick/reindex_v2_job.rb#L16) the soft destroyed record by id, since the default scope won't include it.  Because of this the job [removes](https://github.com/ankane/searchkick/blob/master/lib/searchkick/reindex_v2_job.rb#L30) the record from Esearch.

Hopefully that makes more sense.

ankane:
Ah, gotcha. It probably makes sense to use `unscoped`, but this would be a breaking change, so best to add it as an option for now.

esiegel:
Sounds good to me

afunction:
My problem is same as @esiegel, I use soft_deletion on my project, and theses deleted records can be searched on administrator panel, I change method `self.load_records` on file `lib/searchkick.rb`:

Can I do that? any problem will happened?

```
# private
  def self.load_records(records, ids)
    records =
      if records.respond_to?(:primary_key)
        # ActiveRecord (ADD unscoped)
        records.unscoped.where(records.primary_key => ids) if records.primary_key
      elsif records.respond_to?(:queryable)
        # Mongoid 3+
        records.queryable.for_ids(ids)
      elsif records.respond_to?(:unscoped) && :id.respond_to?(:in)
        # Nobrainer
        records.unscoped.where(:id.in => ids)
      elsif records.respond_to?(:key_column_names)
        records.where(records.key_column_names.first => ids)
      end

    raise Searchkick::Error, "Not sure how to load records" if !records

[...]

PR #1043, opened 2017-12-14
Add unscoped_reindex option that is used during async reindexing.

esiegel:
This adds a new option to the searchkick model `unscoped_reindex` that is used by the ReindexV2Job to unscope the database query.

I wasn't sure the best way to test this?  Do I need to create a Model and give it a default_scope?  Is there some way to do this outside the Model with meta magic?

Issue here: https://github.com/ankane/searchkick/issues/893

### DIFF
--- a/lib/searchkick/model.rb
+++ b/lib/searchkick/model.rb
@@ -5,7 +5,7 @@ module Searchkick
         :filterable, :geo_shape, :highlight, :ignore_above, :index_name, :index_prefix, :inheritance, :language,
         :locations, :mappings, :match, :merge_mappings, :routing, :searchable, :settings, :similarity,
         :special_characters, :stem_conversions, :suggest, :synonyms, :text_end,
-        :text_middle, :text_start, :word, :wordnet, :word_end, :word_middle, :word_start]
+        :text_middle, :text_start, :unscoped_reindex, :word, :wordnet, :word_end, :word_middle, :word_start]
       raise ArgumentError, "unknown keywords: #{unknown_keywords.join(", ")}" if unknown_keywords.any?
 
       raise "Only call searchkick once per model" if respond_to?(:searchkick_index)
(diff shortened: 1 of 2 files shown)


ankane:
Thanks @esiegel 👍  I think this one is fine without a test. I'm worried the name could be mistaken for something else. Any other ideas for names?

esiegel:
@ankane totally agree about naming.  Feels like a weird corner case option that is being exposed, but not sure what would be better.

Also, thinking a little bit more, won't Queued/bulk ids have the same problem? Hmm...

Maybe `:remove_record_scoping`?

ankane:
Thanks @esiegel 👍 

Queued and bulk ids use the `search_import` scope, so they won't have the same issue. One idea is to try and use the `where` parts of the `search_import` scope for lookups (only to relax the default scope). However, think that may add unnecessary complexity. Going to rename to `unscope_reindex_job`. For Searchkick 3, we should be able to do just `unscoped` and not have an option. It's probably safe to do now, but just want to be cautious.

MrHubble:
@ankane any chance you could share how we can use this in Searchkick 3? I see that the `unscoped_reindex_job` option is no longer present, so how do we use `unscoped` with `callback: :async`? Thanks.

ankane:
"It just works"

MrHubble:
haha, so no changes are needed to use `unscoped`? It's unscoped by default?

ankane:
https://github.com/ankane/searchkick/blob/master/lib/searchkick/reindex_v2_job.rb

MrHubble:
Got it. By default ActiveRecord objects will respond to `unscoped`, therefore we apply `unscoped`:

```
if model.respond_to?(:unscoped)
  model.unscoped.find(id)
else
  model.find(id)
end
```

Thanks for this. It was very helpful for me.

PR #1044, opened 2017-12-14
Update README `keyword` usage

divineforest:
to the current ES version according to https://github.com/ankane/searchkick/blob/61ede188042c843c1dfd15fc724379ce69353887/test/test_helper.rb#L457

Otherwise, it's confusing.


ankane:
Great, thanks @divineforest 👍

divineforest:
@ankane thank you for the gem! 👍

PR #1045, opened 2017-12-20
Add references and model_references options to Query and Results

lannon:
We can currently specify `includes` and `model_includes` options when generating a Query, which end up as combined `ActiveRecord::Relation#includes` arguments when generating a results query. 

However, if the query references an included association, an `ActiveRecord::StatementInvalid` error will be generated unless `ActiveRecord::Relation#references` is used to tell ActiveRecord to use a JOIN. 

This PR whitelists `references` and `model_references` keywords in the `Query` class and passes them when constructing an instance of `Results`. The `Results` class then conditionally calls `ActiveRecord::Relation#references` on the query relation.

### DIFF
--- a/test/sql_test.rb
+++ b/test/sql_test.rb
@@ -196,6 +196,12 @@ class SqlTest < Minitest::Test
     assert Product.search("product", includes: [:store]).first.association(:store).loaded?
   end
 
+  def test_references
+    skip unless defined?(ActiveRecord)
+    store_names ["Product A"]
+    assert Product.search("product", includes: [:store], references: [:store]).first.association(:store).loaded?
+  end
+
   def test_model_includes
     skip unless defined?(ActiveRecord)
 
@@ -204,6 +210,25 @@ class SqlTest < Minitest::Test
 
     associations = {Product => [:store], Store => [:products]}
     result = Searchkick.search("*", index_name: [Product, Store], model_includes: associations)
+
+    assert_equal 2, result.length
+
+    result.group_by(&:class).each_pair do |klass, records|
+      assert records.first.association(associations[klass].first).loaded?
+    end
+  end
+
+  def test_model_references
+    skip unless defined?(ActiveRecord)
+
+    store_names ["Product A"]
+    store_names ["Store A"], Store
+
+    associations = {Product => [:store], Store => [:products]}
+    result = Searchkick.search("*",
+                               index_name: [Product, Store],
+                               model_includes: associations,
+                               model_references: associations)
 
     assert_equal 2, result.length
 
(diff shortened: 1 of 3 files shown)


test/sql_test.rb
```diff
@@ -196,6 +196,12 @@ def test_includes
     assert Product.search("product", includes: [:store]).first.association(:store).loaded?
   end
 
+  def test_references
+    skip unless defined?(ActiveRecord)
+    store_names ["Product A"]
+    assert Product.search("product", includes: [:store], references: [:store]).first.association(:store).loaded?
+  end
```
lannon: Wasn't sure how to test that specifying `references` produces the appropriate JOIN clause, but did want to demonstrate that specification of `references` does not introduce regressions with respect to `includes`.

test/sql_test.rb
```diff
@@ -211,4 +217,23 @@ def test_model_includes
...
+                               index_name: [Product, Store],
+                               model_includes: associations,
+                               model_references: associations)
+
+    assert_equal 2, result.length
+
+    result.group_by(&:class).each_pair do |klass, records|
+      assert records.first.association(associations[klass].first).loaded?
+    end
+  end
```
lannon: Same as above. Not sure how to explicitly test the presence of the JOIN clause.

lannon:
The CI failures seem to be order dependent. I can reproduce with `SEED=38027 rake test`, but the error does not seem directly related to the code changes here. It *seems* like it might be related to a lag with `Index#refresh`. If I add a `sleep 0.8` to the failing test, it passes.

lannon:
I can also reproduce this failure on master by adding a random test to `test/sql_test.rb` that simply sleeps for 1s and running the suite with `SEED=38027 rake test`.  Since this failing test only refreshes the Product index, I'm thinking this if this related to searching without specifying an index and relying on `_all`.

ankane:
Hey @lannon, thanks for the PR, including tests 💯  Don't worry about the failures - it's an issue I'm having trouble tracking down.

Can you explain more about what situations `references` is needed? I understand how it's typically used in Rails, but don't see how you would join on the Searchkick model (outside using the `records` method, which will go away in Searchkick 3).

lannon:
Thanks for the feedback, @ankane. I'll put together a test case that demonstrates an SQL error when references is not used. I ran into this while porting some search functionality that was using ActiveRecord to use Searchkick. The case I encountered has some rather complex associations.

It definitely would only be relevant in the context of the `Results#records` method, or more specifically, the `Results#results_query`. If those are going away, this wouldn't be relevant in Searchkick 3. Although, it does seem useful for v2.

PR #1054, opened 2018-01-05
Added an option to return all highlights

unenemy:
Let's give developers option to get all highlights instead of only first one. The default behavior remains, so existing user's code won't be broken by unexpected array instead of string returns.
Tests written, documentation extended.

### DIFF
--- a/README.md
+++ b/README.md
@@ -848,6 +848,22 @@ bands.each do |band|
 end
 ```
 
+**Note:** If you have multiple highlights per field (whether you have an array field or multiple highlighted fragments) searchkick will return only the first one by default.
+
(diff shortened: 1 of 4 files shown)


ankane:
Hey @unenemy, thanks for the PR 👍 I didn't realize it wasn't highlighting the entire field, which is why only one fragment was returned. Searchkick 3 introduces a new way to get highlights - the `with_highlights` method - which takes a `multiple: true` argument for this.

cars10:
Is there any way to set that option before searching?

We do not use the `with_highlights` method in the view, we show highlights by directly accessing `model.search_highlights`

PR #1059, opened 2018-01-10
Closes #1057 updates readme to clarify use of default_fields

### LINKED ISSUE #1057: Documentation References Many Queries That Rely On '_all'
Since Elasticsearch 6 disables the `_all` field by default, should all the queries in the documentation make `fields` explicit?

Also, how would you feel about having default `fields` optionally set on each model?

weilandia:
Clarifies the use of default_fields on a model instead of explicitly setting fields in the query. This helps with the adoption of ES6 as it will no longer default fields to _all.

Closes #1057

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,19 @@ products.each do |product|
 end
 ```
 
+Or
+
+```ruby
+class Product < ApplicationRecord
+  searchkick default_fields: [:name]
+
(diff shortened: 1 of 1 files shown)


ankane:
Hey @weilandia, thanks for the PR 👍 This is no longer needed in the latest version.

PR #1093, opened 2018-03-14
automatically capitalize ENV["CLASS"] in rake task for better UX

raquelxmoss:
Hi folks 👋

Thanks for your work on Searchkick. As this is a very small change, I haven't opened an issue. This PR is just a suggestion, so feel free to close if it you're not keen on it.

While pairing with a co-worker, his natural instinct when running the `reindex` rake task was to pass the model name as lowercase (`class=product` rather than `class=Product`). This was rejected with an error message that explained how to use the CLI.

I thought it might be nicer to tolerate both uppercase and lowercase model names, as long as they are spelt correctly 😸. This is consistent with Rails's rake tasks, and removes a bit of friction for the developer.

### DIFF
--- a/lib/searchkick/tasks.rb
+++ b/lib/searchkick/tasks.rb
@@ -2,7 +2,7 @@ namespace :searchkick do
   desc "reindex model"
   task reindex: :environment do
     if ENV["CLASS"]
-      klass = ENV["CLASS"].constantize rescue nil
+      klass = ENV["CLASS"].capitalize.constantize rescue nil
       if klass
         klass.reindex
       else


lib/searchkick/tasks.rb
```diff
@@ -2,7 +2,7 @@
   desc "reindex model"
   task reindex: :environment do
     if ENV["CLASS"]
-      klass = ENV["CLASS"].constantize rescue nil
+      klass = ENV["CLASS"].capitalize.constantize rescue nil
```
mikelkew: Perhaps `classify` may be even more useful here than `capitalize`?

raquelxmoss: It depends 🙂I avoided using `classify` so as to support non-Rails apps, but if Searchkick is happily Rails-only then I can change it.

mikelkew: Sure, good point! Forgot that it was a Rails only method.

ankane:
Hey @raquelxmoss, thanks for the PR 👍 I think this will have unintended side effects for class names like `ATM` (which will be translated to `Atm`). We could use `classify` (in cases where it's defined), but I'm worried this could also have the same issue with other names.

raquelxmoss:
@ankane Ah you're right! Didn't think about that case, d'oh!

PR #1098, opened 2018-03-15
Fix infinite recursion on multi search

fidalgo:
When we retry a multi search, we should not perform subsequent retries,
otherwise we'll end in an inifite recursion if we don't have enough
matches.

In this commit, a new flag is added to mark the execution of a retry.
Also since we have a local variable and a method with the same name
`queries`, a rename was applied to the method argument to improve the
readability.

### DIFF
--- a/lib/searchkick/multi_search.rb
+++ b/lib/searchkick/multi_search.rb
@@ -14,12 +14,12 @@ module Searchkick
 
     private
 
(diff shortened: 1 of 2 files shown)


ankane:
Looks great, thanks @fidalgo 💯 (extra thanks for updating the test)

PR #1099, opened 2018-03-15
Add the frozen string literal

fidalgo:
The adding of the frozen literal string will reduce the amount of
memory needed due the String imutabilty.
Works only on Ruby 2.3+, ignored on others.


ankane:
Hey @fidalgo, thanks for another PR 👍 I tried this a few weeks ago and there was pretty much no memory impact.

PR #1132, opened 2018-04-30
Remove duplicated rake require

y-yagi:
There seems to be a duplicate process related to Rake.
https://github.com/ankane/searchkick/blob/d7312a300e56c035dcbd217c27438c6ed5a6484a/lib/searchkick.rb#L22-L34


ankane:
Thanks @y-yagi 👍

PR #1134, opened 2018-05-02
Avoid loading the Records in memory in async mode

fidalgo:
Since in async mode we only need the IDs for the job, we can avoid to
load the database records.

We'll only loaded when and if we need need them.

### DIFF
--- a/lib/searchkick/bulk_indexer.rb
+++ b/lib/searchkick/bulk_indexer.rb
@@ -14,7 +14,7 @@ module Searchkick
       end
 
       if batch
-        import_or_update relation.to_a, method_name, async
+        import_or_update relation, method_name, async
         Searchkick.with_redis { |r| r.srem(batches_key, batch_id) } if batch_id
       elsif full && async
         full_reindex_async(relation)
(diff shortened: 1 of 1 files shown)


lib/searchkick/bulk_indexer.rb
```diff
@@ -56,27 +56,24 @@ def batches_left
...
-            with_retries do
-              # call out to index for ActiveSupport notifications
-              if method_name
-                index.bulk_update(records, method_name)
-              else
-                index.bulk_index(records)
-              end
-            end
+      return if records.none?
+      if async
```
ankane: Please keep change set to a minimum (aka don't unnecessarily change indentation)

lib/searchkick/bulk_indexer.rb
```diff
@@ -56,27 +56,24 @@ def batches_left
...
-          if records.any?
-            with_retries do
-              # call out to index for ActiveSupport notifications
-              if method_name
-                index.bulk_update(records, method_name)
-              else
-                index.bulk_index(records)
-              end
-            end
+      return if records.none?
```
ankane: This will cause an additional query. I think a better approach would be to call `records.to_a` or `pluck(:id)` first, then check if `any?`.

lib/searchkick/bulk_indexer.rb
```diff
@@ -56,27 +56,24 @@ def batches_left
...
+      if async
+        Searchkick::BulkReindexJob.perform_later(
+          class_name: records.first.class.name,
```
jduarte: I think that here we should use `records.klass.name` and that would avoid that extra query

lib/searchkick/bulk_indexer.rb
jduarte: `.to_a` will load all the AR objects into memory that may cause a bloat if we're talking on a considerable amount of AR objects so IMO I think it's best to avoid that.

Although not so pretty I think that the best option here is something like:

```ruby
if async
  record_ids = records.pluck("#{records.klass.table_name}.id")
  return if record_ids.empty?
  Searchkick::BulkReindexJob.perform_later(
    class_name: records.klass.name,
    record_ids: records.pluck(:id),
    index_name: index.name,
    method_name: method_name ? method_name.to_s : nil
  )
else
  loaded_records = records.to_a.select(&:should_index?)
  return if loaded_records.size.zero?
  # ...
end
```

jduarte:
Left some insights as well :) Hopeful to see this merged

lib/searchkick/bulk_indexer.rb
ankane: Agreed

lib/searchkick/bulk_indexer.rb
ankane: That's exactly what I mean above. There's an easy to do this with a minimal change set.

ankane:
Hey @fidalgo, as far as I can tell, the combination of `async: true` and `batch: true` is never used with `import_scope`. Please let me know if I'm missing something.

PR #1136, opened 2018-05-05
Add time zone to date histogram

geeosh:
I was working on an aggregation view by day, where the day buckets were relative to the user's time zone. I discovered Searchkick didn't support the [Elasticsearch feature](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#_time_zone_2) where you can pass in a `time_zone` argument when doing aggregations.

I've added support for it, and I also added a test for the usage of the time zone attribute. Hope this can help others!

Let me know if you'd like me to clean up the test or implementation in any way.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -715,11 +715,12 @@ module Searchkick
           }
         elsif (histogram = agg_options[:date_histogram])
           interval = histogram[:interval]
+          time_zone = histogram[:time_zone].present? ? { time_zone: histogram[:time_zone] } : {}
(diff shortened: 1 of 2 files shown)


geeosh:
@ankane thanks for the suggestions!

1. I updated the aggregation filter so it passed in all options in the `agg_options[:date_histogram]` hash. Let me know if we need any parameter sanitization.

2. I've updated the tests so they are not affected by the time of day. Let me know if you think they're adequate.

geeosh:
@ankane 🙌

Issue #164, opened 2014-03-30
Search with misspellings on certain field.


andreikarabitski:
Hello, 
do I have a possibility to search on several fields defined in index with misspellings disabled, but certain field with enabled misspellings? I guess I have to create custom mapping for that field?

sweetmandm:
Here's my hacky workaround in the meantime. Heads-up that this is deadline-mode-code 😂 It's probably specific to the queries I'm making. Thanks for pointing `execute: false` & `query.body`!

I haven't really grokked the elasticsearch docs but I noticed that (with mostly-default searchkick queries anyway) leaving misspellings enabled results in two extra query body entries vs. adding `misspellings: false`. This code just tries to remove the two extra entries that look like they were added to handle misspellings, and it seems to be working for me.
```
  def run
    q = Product.search("Test", fields: [:title, :description], execute: false)
    disable_misspellings!(q.body, ["title.analyzed"])
    q.execute
  end

  def disable_misspellings!(query_body, keys)
    return unless query_body.dig(:query, :dis_max, :queries)

    # Remove any query containing at least one of misspelling_params
    misspelling_params = [
      :fuzziness,
      :prefix_length,
      :max_expansions,
      :fuzzy_transpositions
    ]

    keys.each do |key|
      items = query_body[:query][:dis_max][:queries]
      query_body[:query][:dis_max][:queries] = items.reduce([]) do |acc, item|
        match_params = item.dig(:match, key).try(:keys) || []
        if (match_params & misspelling_params).any?
          # This looks like a misspelling query! don't include it in 'acc'
          acc
        else
          acc.append(item)
        end
      end
    end
  end
```

PR #1137, opened 2018-05-06
Per-field Misspelling Options

sweetmandm:
## Why?
https://github.com/ankane/searchkick/issues/164
It can sometimes be helpful to control misspellings at a per-field level. The main use case I have in mind is for domains where items are commonly referred to either by name or by an ID code.

Please see the [caveats](#caveats) section below.

## What Changed?
This does not break existing API. It changes the `misspellings` interface to allow specifying `fields`. When specifying fields for misspellings, it is also necessary to specify which fields to search. The defaults and root-level misspelling options still apply, but the per-field options override those defaults. For example, the following will search `name` and `color` but will disable misspellings in `color`, while allowing them in `name`:
```ruby
Product.search "zucini",
  fields: [:name, :color],
  misspellings: {fields: {color: false}}

# Note: Updated syntax as of 3d23519 only allows turning fields on/off:
Product.search "zucini",
  fields: [:name, :color],
  misspellings: {fields: [:color]}
```

The existing options are still available on the `misspellings` hash and they become the default for any fields that are not stated explicitly. For example, to only allow transpositions on `color`:
```ruby
Product.search "zuccihni",
  fields: [:name, :color],
  misspellings: {
    transpositions: false,
    fields: {color: {transpositions: true}}
  }

# Note: Updated syntax as of 3d23519 does not support merging options per-field, it only allows enabling or disabling on a per-field basis
```

`below` is the only option not available per-field.

<a name="caveats"></a>
## Caveats
I dislike the interface complexity this adds to the `misspellings` option, specifically the "cascading option" effect and the requirement to specify all `fields` to search. I do not know if the benefit vs. complexity cost is worth it. It's always possible to modify the query as described in #164, which is a more DIY approach that could cause less confusion around the interface. Maybe somebody will see a way to simplify/clarify the interface.

Thanks for searchkick!


sweetmandm:
I don't think I can see the Travis failure reason, but I bet it's the use of the safe navigation operator, which came in ruby 2.3.0 -- just realized it looks like [Searchkick should work with >= 2.2.0](https://github.com/ankane/searchkick/blob/master/searchkick.gemspec#L21)

sweetmandm:
🤔 The failure is in an `edit_distance` test, and only fails on ES 5.0.1. I notice that `edit_distance` does not appear in the rest of the tests, other than to test the default behavior. I wonder if there's some quirk with `edit_distance` & 5.0.1?

sweetmandm:
Aha, looks like 5.0.1 didn't handle edit_distance with a two-character query string (`bl` matching `blue` -- seems fine with four characters (`yell` matching `yellow`)

lib/searchkick/query.rb
```diff
@@ -490,10 +492,26 @@ def set_fields
           []
         elsif default_match == :exact
           raise ArgumentError, "Must specify fields to search"
+        elsif has_field_misspellings(options[:misspellings])
+          raise ArgumentError, "If you provide per-field misspelling rules you must also provide the fields to search."
```
ankane: Let's do the `ArgumentError` check in `prepare` so we don't pollute `set_fields`.

lib/searchkick/query.rb
```diff
@@ -490,10 +492,26 @@ def set_fields
...
+
+    def has_field_misspellings(misspellings)
+      misspellings.is_a?(Hash) &&
+        misspellings[:fields].is_a?(Hash) &&
+        misspellings[:fields].keys.any?
+    end
+
+    def misspellings_for_field(misspellings, field)
+      if misspellings.is_a?(Hash) && misspellings[:fields] && misspellings[:fields][field] != nil
+        misspellings[:fields][field]
```
ankane: Think we should merge this with `misspellings` if it's a hash, so `misspellings: {edit_distance: 2, fields: {color: {transpositions: false}}` will keep the edit distance for color.

test/per_field_misspellings_test.rb
```diff
@@ -0,0 +1,147 @@
...
+  end
+
+  def test_misspelling_zucchini_transposition
+    store [{name: "zucchini", color: "green"}]
+    assert_search "zuccihni", [],
+      fields: [:name, :color],
+      misspellings: { fields: { name: {transpositions: false} } }
+    assert_search "grene", ["zucchini"],
+      fields: [:name, :color],
+      misspellings: { fields: { name: {transpositions: false} } }
```
ankane: Would be good to have a helper method like `assert_misspellings` to DRY the tests. Also, let's add to existing misspellings test instead of a new file and use common prefix for method names, like `test_misspellings_field_...`.

ankane:
Another simpler (but less flexible) option would be to add a `misspellings: {fields: [:color]}` if the primary use case will be only having misspellings apply to some fields.

sweetmandm:
Thanks for the feedback @ankane! Just to double-check -- do you prefer this PR's current interface or the `{fields: [:color]}` style? I would be happy with either, and I'd want to use whichever feels more aligned with searchkick's general philosophy

lib/searchkick/query.rb
sweetmandm: :+1: updated in 15cc52b58fe0c7127856381d712e78d8c08befb0

test/per_field_misspellings_test.rb
sweetmandm: updated in https://github.com/ankane/searchkick/pull/1137/commits/28f1fb2da2fdc811351e80916e5b71f799cd21f6 (for `assert_misspellings`)
and https://github.com/ankane/searchkick/pull/1137/commits/29933aac41e6b0831830d371fd83a7658427a21c (combining the two files)

lib/searchkick/query.rb
sweetmandm: updated in 9e7a7e1101a6b1f241ee9332177e5889a44c0f24

ankane:
Hey @sweetmandm, sorry for the delay. I think it'd be better to start with the simpler `fields: [:color]` option and see how that looks. I feel like the code and test cases add a lot of complexity in the current form. Also, not sure I understand why a change in `set_fields` is needed.

sweetmandm:
Thanks @ankane, agreed, I'll switch it up to use `fields: [:color]`.

There may be a way to accomplish this without touching `set_fields` -- but it seemed the best way to be certain the field name translations were identical. In other words, since `set_fields` translates the from e.g. `['color^2']` to `['color.analyzed']`, I needed a way to match the `:color` symbol that would show up in misspelling options against the translated versions produced by `set_fields`. For this example assuming a field named `color`, the resulting `option_fields_mapping` would look like `{'color.analyzed': :color}`.

Since [fields.each](https://github.com/ankane/searchkick/pull/1137/files#diff-9e535cec87dc0b7137c4f81bdabdafd7R277) iterates over the `'color.analyzed'` version, it [looks into option_fields_mapping, pulling out `:color`](https://github.com/ankane/searchkick/pull/1137/files#diff-9e535cec87dc0b7137c4f81bdabdafd7R306), and then looks at the misspelling options for `:color` to determine if misspellings should be applied.

ankane:
Thanks @sweetmandm 👍 Did a little refactoring and merged in https://github.com/ankane/searchkick/commit/31f846a08c8bb84a301c62217d6b12ff11dd6141

Issue #1151, opened 2018-05-17
Force Reindex with Record Ids instead of Min/Max


earoberson:
Hey @ankane 

I'm trying to reindex a model that has a non-sequential primary key `id`, but the primary key _is_ numeric.

I'm looking at: https://github.com/ankane/searchkick/commit/58af0f694baeca9e2e6b28a758c950894dc9078d

The issue I'm facing is that the number of jobs created when async reindexing becomes extremely large and often doesn't cover more than a handful of actual records since a range of 1000 sequential ints could potentially cover only 2-3 ids. 

Is there a way to force creating a reindex job with record_ids instead of a min and max id?

Right now, the hack is to raise an ActiveRecord::StatementInvalid error in `minimum` and everything works well.

ankane:
Closing so discussion is only in one place (PR)

PR #1152, opened 2018-05-17
Batch with record_ids instead of bounds

earoberson:
When trying to reindex a model that has a non-sequential primary key `id`, but the primary key _is_ numeric, there is an issue when async reindexing - the number of jobs becomes extremely large and often doesn't cover more than a handful of actual ids. 

This feature allows optionally using record_ids instead of min_id and max_id in the async job, even if the scope's primary id is numeric.

The default is to use existing functionality.

For reference: https://github.com/ankane/searchkick/issues/1151

### DIFF
--- a/lib/searchkick/bulk_indexer.rb
+++ b/lib/searchkick/bulk_indexer.rb
@@ -86,13 +86,13 @@ module Searchkick
       if scope.respond_to?(:primary_key)
         # TODO expire Redis key
         primary_key = scope.primary_key
+        primary_key_type = scope.columns_hash[primary_key].type
 
-        starting_id =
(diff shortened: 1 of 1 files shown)


ankane:
Hey @earoberson, thanks for the PR 👍 A few comments:

- Let's keep the same logic when the new option isn't specified (rather than checking the column type)
- Let's nest the option under the `async` option
- The option name is super clear to me

ankane:
Cleaning up stale PRs

ankane:
Hey @earoberson, just fyi, this now how async reindexing works in Searchkick 5. Thanks again for the PR.

PR #1161, opened 2018-06-07
fix: prevent generate duplicate field from suggest field (string/symbol)

wokenqingtian:
Issue:
When the value of **suggest filed is nill**, searchkick reindex task will create two duplicate fields in query parameters. And also will get an error from Elasticsearch below:

```
Searchkick::ImportError: {"type"=>"mapper_parsing_exception", "reason"=>"failed to parse", "caused_by"=>{"type"=>"json_parse_exception", "reason"=>"Duplicate field 'suggest_field'\n at [Source: org.elasticsearch.common.bytes.BytesReference$MarkSupportingStreamInputWrapper@62b97d8e; line: 1, column: 163]"}} on item with id '653'
```

Example:

```ruby
class Person < ActiveRecord::Base
  searchkick suggest: [:suggest_field]

  def search_data
    {
      name: name,
      suggest_field: suggest_field
    }
  end
end
```
(The suggest_filed's value in database is NULL )
When run the reindex task, the parameters in the query will like this:

```
 {:index=>
   {:_index=>"locales_development_20180607231727213",
    :_id=>653,
    :_type=>"person",
    :data=>
     {:name=>nil,
      :suggest_field=>nil,
      "suggest_field"=>nil}}},
```
You can find two suggest_field in the hash. One is string, another is symbol. 
The string is generate by the method `search_data` in RecordData.

My PR is to fix this duplicate issue.

### DIFF
--- a/lib/searchkick/record_data.rb
+++ b/lib/searchkick/record_data.rb
@@ -61,8 +61,8 @@ module Searchkick
       # hack to prevent generator field doesn't exist error
       if !partial_reindex
         index.suggest_fields.each do |field|
-          if !source[field] && !source[field.to_sym]
-            source[field] = nil
+          if (source.keys & [field, field.to_sym]).empty?
+	          source[field] = nil
           end
         end
       end


ankane:
Hey @wokenqingtian, thanks for the PR 👍 It might be my brain today, but I don't see how the proposed change is logically different than the current code.

wokenqingtian:
@ankane 
We can consider two exmaples and do some comparison on them.

Example 1:
```ruby
class Person < ActiveRecord::Base
  searchkick suggest: [:suggest_field]

  def search_data
    {
      name: name,
      suggest_field: suggest_field
    }
  end
end
```
And searchkick get a record:

```
{ name: "petter", suggest_feild: nil }
```

Exampe 2:

```ruby
class Person < ActiveRecord::Base
  searchkick suggest: [:suggest_field]

  def search_data
    {
      name: name,
    }
  end
end
```
And searchkick get a record below:
```
{ name: "peter" }
```

Then we use the original approach to handle these two examples:

```ruby
index.suggest_fields.each do |field|
  if !source[field] && !source[field.to_sym]
    source[field] = nil
  end
end
```

* The result of example 1: { :name => "petter", :suggest_field => nil, "suggest_field" => nil }
* The result of example 2: { :name => "petter", "suggest_field" => nil }

See, the original approach inserts same `"suggest_field" => nil` into the hash. It seems that the original approach is designed to handle example two. But when it comes to example 1, the `:suggest_field` we declare in `search_data` is a symbol, and the `souce[filed] = nil` just use a string, then the result will have two duplicated fields (`:suggest_field`, `"suggest_field"`).

So I use `(source.keys & [field, field.to_sym]).empty?` to directly check the keys. Then the two results will be like this:

* The result of example 1: { :name => "petter", :suggest_feild => nil }
* The result of example 2: { :name => "petter", "suggest_field" => nil }

ankane:
Ah, I see. Thanks @wokenqingtian 👍

PR #1162, opened 2018-06-11
Update README.md: add example how to delete one record from the index

shiroginne:
Hello,

thanks for the gem, it helped us a lot!

Recently I had one problem: how to delete one specific record from the index? I didn't find the answer in README, so I started to look into source code and find `Searchkick::Index#remove(record)` which do the job.

I think it would be a good idea to add this trick right in README, so other people will find how to it quicker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1518,6 +1518,13 @@ Reindex associations
 store.products.reindex
 ```
 
+Remove one record from the index
+
+```ruby
+product = Product.find(1)
+Product.search_index.remove(product)
+```
+
 Remove old indices
 
 ```ruby


ankane:
Thanks for the PR @shiroginne 👍 What's the main use case for this?

shiroginne:
We use ES + SQ for special place something like shop-window, and we need to opt out special companies from there if they choose not to be shown. Reindex all records would be bad idea for 1 record.

ankane:
Have you tried the `should_index?` method to accomplish this?

shiroginne:
Yes, but did it remove from the existing index?
For example:
we automatically add all companies, some company decided not to be shown, they choose "invisible" and we have something like:
```ruby
def should_index?
  visible?
end
```

in this case there will be false. Will record be removed from the index?

ankane:
Yes, once you reindex the record if will be (which happens when you save it by default).

shiroginne:
Thanks, didn't know that. Well, for us `should_reindex?` will work then. As for README don't know if it make sense now, maybe could be useful to remove some other records on callbacks 🤷‍♂️

ankane:
The problem with removing manually is they'll be re-added when they're reindexed again, which is why `should_index?` is the recommended method. Going to leave this out of the readme for now to avoid confusion.

Issue #642, opened 2016-03-08
Result window is too large


theonegri:
I'm using this gem and Kaminari for the pagination  to search through roughly 7 millions documents.

When clicking on "Last »" the pagination I get the error: "Result window is too large, from + size must be less than or equal to: [10000] but was [6745575]"

ankane:
This is a limit Elasticsearch places on pagination.  You can change the limit with `index.max_result_window` in Elasticsearch, or set `max_pages` in Kaminari.

milushov:
Hello, @ankane, can you tell how to set `index.max_result_window` through searchkick? I didn't found any references in the readme.

ankane:
It's a setting on the Elasticsearch server, so you aren't able to set it through Searchkick.

milushov:
@ankane thanks for info

for those who will encounter same issue: here is quick workaround:
```
# items_staging - name of the index
# localhost:9200 - default host and port
curl -XPUT "http://localhost:9200/items_staging/_settings" -d '{ "index" : { "max_result_window" : 100000 } }'
# ckecking that setting saved ok
curl -s -XGET "http://localhost:9200/items_staging/_settings?pretty" | grep window
```

ankane:
Didn't realize you could set it with `_settings`. In that case, you should be able to do:

```ruby
class Item < ActiveRecord::Base
  searchkick settings: {index: {max_results_window: 100000}}
end
```

Lewis-Clayton:
Just in case anyone else is copy paste happy like me. The above should be **max_result_window**
```
class Item < ActiveRecord::Base
  searchkick settings: {index: {max_result_window: 100000}}
end
```

ankane:
Nice catch, updated :)

ankane:
Also, be sure you understand to implications of increasing it.

---

To understand why deep paging is problematic, let’s imagine that we are searching within a single index with five primary shards. When we request the first page of results (results 1 to 10), each shard produces its own top 10 results and returns them to the coordinating node, which then sorts all 50 results in order to select the overall top 10.

Now imagine that we ask for page 1,000—results 10,001 to 10,010. Everything works in the same way except that each shard has to produce its top 10,010 results. The coordinating node then sorts through all 50,050 results and discards 50,040 of them!

You can see that, in a distributed system, the cost of sorting results grows exponentially the deeper we page. There is a good reason that web search engines don’t return more than 1,000 results for any query.

[...]

PR #1165, opened 2018-06-19
Elasticsearch Scroll API

kschutt:
## What's up

* We were running into similar problems to #642 but found that simply increasing the result window was not a viable solution

## What this does

* Adds support for [Elasticsearch Scroll API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html)
* Allows users to pass in the `scroll` param to initiate a scroll context
* Stores the `scroll_id` on the `Searchkick::Results` object
* Exposes a `scroll` method on `Searchkick::Results` to request the next batch of results
* `nils` out pagination-related methods if using a scroll context

## Notes

* Since the Scroll API uses a URL param, I wasn't quite sure how to represent that in `Searchkick::Query` so I just followed what was already there for pagination
* I force delete  the `from` attribute in the payload because Scroll Contexts do not allow that parameter
* I wasn't sure of the protocol of using the `Searchkick.client` from within the `Searchkick::Results` object. Is this allowed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,20 @@ View with will_paginate
 <%= will_paginate @products %>
 ```
 
+### Scroll API
+
+For large datasets, a scroll context is more efficient than deep pagination. Create a scroll context by passing in the desired keep-alive time and the number of records per page
+
+```ruby
+products = Product.search "milk", per_page: 10, scroll: '1m'
+```
+
+Request the next set of records in the scroll context
+
+```ruby
+next_records = products.scroll
+``` 
+
 ### Partial Matches
 
 By default, results must match all words in the query.
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -12,14 +12,14 @@ module Searchkick
       :took, :error, :model_name, :entry_name, :total_count, :total_entries,
       :current_page, :per_page, :limit_value, :padding, :total_pages, :num_pages,
       :offset_value, :offset, :previous_page, :prev_page, :next_page, :first_page?, :last_page?,
-      :out_of_range?, :hits, :response, :to_a, :first
+      :out_of_range?, :hits, :response, :to_a, :first, :scroll
 
     def initialize(klass, term = "*", **options)
       unknown_keywords = options.keys - [:aggs, :body, :body_options, :boost,
         :boost_by, :boost_by_distance, :boost_by_recency, :boost_where, :conversions, :conversions_term, :debug, :emoji, :exclude, :execute, :explain,
         :fields, :highlight, :includes, :index_name, :indices_boost, :limit, :load,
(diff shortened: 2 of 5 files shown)


lib/searchkick/results.rb
```diff
@@ -137,46 +137,56 @@ def total_count
     alias_method :total_entries, :total_count
 
     def current_page
+      return if scroll_id
```
ankane: What's the motivation for having pagination methods return `nil`? I think we can keep track of the page even with the scroll method.

lib/searchkick/query.rb
```diff
@@ -460,6 +465,12 @@ def prepare
         payload[:from] = offset
       end
 
+      # scroll
+      if scroll
```
ankane: Let's raise an `ArgumentError` if invalid parameter combinations are passed instead of silently failing.

test/scroll_test.rb
```diff
@@ -0,0 +1,128 @@
+require_relative "test_helper"
+
+class ScrollTest < Minitest::Test
+  def test_limit
```
ankane: Let's remove all tests that don't involve the the scroll API. Existing functionality should already be tested.

lib/searchkick/results.rb
```diff
@@ -204,6 +214,24 @@ def misspellings?
...
+      if scroll_id.present? && options[:scroll].nil?
+        raise Searchkick::Error, "Scroll error - scroll keepalive must be defined"
+      elsif scroll_id.nil?
+        raise Searchkick::Error, "Scroll error - a scroll id has not been provided"
+      else
+        params = {
+          scroll: options[:scroll],
+          scroll_id: scroll_id
+        }
+        Searchkick::Results.new(@klass, Searchkick.client.scroll(params), @options)
```
ankane: Would be great to return a friendly error message when the scroll is expired.

README.md
```diff
@@ -241,6 +241,20 @@ View with will_paginate
...
+For large datasets, a scroll context is more efficient than deep pagination. Create a scroll context by passing in the desired keep-alive time and the number of records per page
+
+```ruby
+products = Product.search "milk", per_page: 10, scroll: '1m'
+```
+
+Request the next set of records in the scroll context
+
+```ruby
+next_records = products.scroll
```
ankane: The docs state:

> The initial search request and each subsequent scroll request each return a _scroll_id, which may change with each request — only the most recent _scroll_id should be used.

So we want the scroll method to be called on the new result set each time instead of the original. I'm not sure how we want to get this across in the docs.

ankane:
Cleaning up stale PRs

ckhall:
@kschutt @ankane any chance of reviving this?

PR #1170, opened 2018-06-29
Fix the total docs call after import logic move

erichkist:
Hi folks,

When I tried to reindex with `resume`, the error raises:
```ruby
NameError: undefined local variable or method `total_docs' for #<Searchkick::BulkIndexer:0x007ff78b9b2168>
```

I believe that the erro was introduced after the import logic move on commit fd936e493ec46909171a849d9f4ed68130ede7ec.

Thanks :)


ankane:
Thanks @erichkist 👍 will add a test for it as well.

PR #1174, opened 2018-07-12
Adds a :multiple option for searching with :highlight for .search_highlights with multiple results

cars10:
Currently you can already retrieve multiple highlights like this:

```ruby
bands = Band.search "cinema", highlight: {fragment_size: 20}
bands.with_highlights(multiple: true).each do |band, highlights|
  highlights[:name].join(" and ")
end
```

Sadly this approach leads to quite complex rendering and view handling (because you cannot directly loop the records but have to loop an array of `[record, highlight]`). 

Because of that i normally use the method `.search_highlights` on `record` that is automatically defined by searchkick. This PR also enables the `multiple` behaviour of `.with_highlights` for the `.search_highlights` method by simply specifying a `multiple` option when searching. Example:

```ruby
Product.search("cinema", highlight: {fragment_size: 20, multiple: true})
```

Using `record.search_highlights` in the view then returns multiple highlights just like `.with_highlights` does.

### DIFF
--- a/lib/searchkick/results.rb
+++ b/lib/searchkick/results.rb
@@ -38,7 +38,8 @@ module Searchkick
               result = results[hit["_type"]][hit["_id"].to_s]
               if result && !(options[:load].is_a?(Hash) && options[:load][:dumpable])
                 if (hit["highlight"] || options[:highlight]) && !result.respond_to?(:search_highlights)
-                  highlights = hit_highlights(hit)
+                  multiple = options[:highlight].is_a?(Hash) && options[:highlight][:multiple]
+                  highlights = hit_highlights(hit, multiple: multiple)
                   result.define_singleton_method(:search_highlights) do
                     highlights
                   end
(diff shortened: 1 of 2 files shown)


ankane:
Hey @cars10, thanks for the PR 👍 Unfortunately, I don't think another option for this makes sense. Adding an extra argument to loops shouldn't be too difficult (unless I'm missing something).

cars10:
Hi, you are right - its not difficult. But you cannot use features like collection rendering with the current structure. WIth my current approach i can simply do 

```ruby
<%= render resuls %>
```

or

```ruby
<%= render collection: results %>
```

and thus use caching and collection rendering features of rails. Wiht your proposed approach i cannot use these. Apart from that - why do you even provide the `.search_highlights` method if `.with_highlights` should be the way to go?

ankane:
`search_highlights` is a convenience method if you don't need multiple fragments. You should be able to pass the `with_highlights` result to partials, as it's still enumerable.

```erb
<%= render partial: "product", collection: result.with_highlights(multiple: true) %>
```

Not sure I follow the caching part.

ankane:
In general, "why do you even provide the X feature if it doesn't fit my need?" is not a way to win friends in open source.

PR #1205, opened 2018-11-09
add description to gemspec

lostapathy:
This adds a short description to the gemspec.  

Specifically, I wanted to add the word "Elasticsearch" so that this gem is more discoverable on RubyGems by potential users who need an Elasticsearch client.

Obviously just my first draft of the text trying to make the gem easier to discover, please improve as you see fit.


ankane:
Hey @lostapathy, great idea 👍 Can’t believe I missed that (facepalm).

I usually keep the gem summary and GitHub description synced, so how about adding “with Rails and Elasticsearch” to the summary?

PR #1226, opened 2019-01-30
Implements simple_query_string option

weilandia:
I'm guessing this may be a naive approach to implementing this feature, but thought I'd give it a shot. I know we can use the `body` option to do more advanced queries, including `simple_query_string` queries, but it would be nice to have a way to do it without having to also manage where clauses, pagination, timeouts etc.

I thought another solution to this might be to have a `query_options` option that takes a proc that can inject a query into the other options, but I just needed `simple_query_string` so I went for that.

Let me know what ya'll think!

[implements-simple-query-string]


ankane:
Hey @weilandia, thanks for the PR 👍 I'd prefer not to add another pattern for searching unless it's significantly better than the existing one for common use cases. Check out the Advanced Search section of the readme for how to customize a Searchkick-generated query - I believe it's similar to your `query_options` proposal.

weilandia:
@ankane thanks for your response.  The problem I ran into with Searchkick's Advanced Search is that I then have to also have to use the Elasticsearch dsl for where clauses, pagination, etc.  It would be nice (specifically for the need to use `simple_query_string`) to be able to just inject the query parameter while still using Searchkick for the other options.

ankane:
What you're asking for is already possible.

PR #1235, opened 2019-02-26
Automatically requeue bulk rejections

stormsilver:
In a busy ES cluster with lots of write activity, using the `bulk` callback method, it is highly likely to see lots of these guys:

```
Searchkick::ImportError: {"type"=>"es_rejected_execution_exception", "reason"=>"rejected execution of org.elasticsearch.transport.TransportService$4@f1e9eea on EsThreadPoolExecutor[bulk, queue capacity = 50, org.elasticsearch.common.util.concurrent.EsThreadPoolExecutor@46cb2258[Running, pool size = 2, active threads = 2, queued tasks = 50, completed tasks = 1234]]"} on item with id '1234'
```

When a bulk rejection of this kind occurs, it is possible that _some_ of the operations succeeded and _some_ of them failed. According to the docs on [thread pools](https://www.elastic.co/guide/en/elasticsearch/guide/current/_monitoring_individual_nodes.html#_threadpool_section), what we'd like to do is check the response from ES and retry just the failures.

Prior to this change, the job would raise an exception and attempt to index 100% of the items that were part of the job, regardless of initial success.

After this change, the items that succeeded are considered done and the items that failed are added back to the bulk queue to get picked up the next time a `ProcessQueueJob` runs.

### DIFF
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -32,7 +32,9 @@ module Searchkick
   class UnsupportedVersionError < Error; end
   class InvalidQueryError < Elasticsearch::Transport::Transport::Errors::BadRequest; end
   class DangerousOperation < Error; end
-  class ImportError < Error; end
+  class ImportError < Error
+    attr_accessor :failures
+  end
 
   class << self
     attr_accessor :search_method_name, :wordnet_path, :timeout, :models, :client_options, :redis, :index_prefix, :index_suffix, :queue_name, :model_options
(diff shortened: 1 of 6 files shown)


ankane:
Hey @stormsilver, thanks for the PR 👍 I appreciate the great description. The changeset seems like it could be smaller. What if we add an option to `Searchkick.callbacks` to not raise an exception on errors?

ankane:
Also, thinking about it more, it seems like this could easily lead to records not being indexed properly if there's a mistake in the logic, and since there's not a great way to test it (without stubbing it), I'm a bit hesitant to add complexity to this codepath.

ankane:
Cleaning up PRs

PR #1253, opened 2019-04-12
support include_type_name

ogidow:
Elasticsearch 6.7 adds the include_type_name option for migration to v7.0.
Searchkick also supports include_type_name for smooth migration

https://www.elastic.co/guide/en/elasticsearch/reference/master/removal-of-types.html

### DIFF
--- a/lib/searchkick/index.rb
+++ b/lib/searchkick/index.rb
@@ -13,7 +13,12 @@ module Searchkick
     end
 
     def create(body = {})
-      client.indices.create index: name, body: body
+      include_type_name = body.delete(:include_type_name)
+      if include_type_name
+        client.indices.create index: name, body: body, include_type_name: include_type_name
+      else
+        client.indices.create index: name, body: body
+      end
     end
 
     def delete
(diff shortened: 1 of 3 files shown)


ankane:
Hey @ogidow, thanks for the PR 👍 I looked into that option but found it wasn't needed for a zero-downtime upgrade.

ogidow:
@ankane thanks for your response

Is the method you are thinking that is shown below? 


- In 6.x, create index with _doc type
- upgrade to 7.0
- In 7.0 does not include type in mapping reindex

ankane:
From my tests, you don't need to reindex with `_doc` type at all. You can upgrade to ES 7 while on Searchkick 3 or 4 without downtime. Before you do a full reindex on ES7, you need to upgrade to Searchkick 4 (and make sure it's running on all app servers that reindex).

To keep things simple, here's the approach I'm suggesting in the readme: https://github.com/ankane/searchkick#elasticsearch-6-to-7-upgrade

ankane:
That being said, I'd definitely test it out in staging before trying it in production (if you have a lot of data). Would be good to confirmation that this approach works for others.

ogidow:
@ankane 
[It](https://github.com/ankane/searchkick#elasticsearch-6-to-7-upgrade) seemed that you could upgrade in this way.
thank you for your suggestion

ankane:
Great, no problem

PR #1263, opened 2019-05-23
Update Readme.md Heroku / foundelasticsearch

nicovak:
For Heroku with plugin foundelasticsearch and last version.
I did the update today and got `Faraday::ConnectionFailed - SSL connect error` for each request after the deploy.
I don't know what changed in any dependency or even in searchkick but adding the port resolve the issue. It was working fine with earlier versions of`elasticsearch-ruby` and deps.

Maybe there is the same issue with bonsai addon, i don't know. Maybe someone should verify.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1038,16 +1038,19 @@ heroku addons:create foundelasticsearch
 heroku addons:open foundelasticsearch
 ```
 
-Visit the Shield page and reset your password. You’ll need to add the username and password to your url. Get the existing url with:
+Visit the Shield page and reset your password. You’ll need to add the username and password to your url.
+Also, verify your port number in your dashboard, It sould be `9243`
+
+Get the existing url with:
 
 ```sh
 heroku config:get FOUNDELASTICSEARCH_URL
 ```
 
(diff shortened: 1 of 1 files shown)


ankane:
Hey @nicovak, thanks for the PR. I believe Found was using the standard port before (9200), which is why the port wasn't previously needed. I'd be good to report this issue to Found support (the FOUNDELASTICSEARCH_URL is missing the port), as I bet others who don't use Searchkick are running into the same issue: https://devcenter.heroku.com/articles/foundelasticsearch#support

ankane:
Looks like 9243 is "strongly recommended" going forward: https://www.elastic.co/guide/en/cloud/current/ec-getting-started-connect.html

ankane:
Here's what changed: https://github.com/elastic/elasticsearch-ruby/issues/669

nicovak:
Hi @ankane thanks for the link, I was pretty sure that It was coming from `elasticsearch-ruby`. I didn't find the issue when I was looking for It. I noticed that `9243` is the recommended port too.

PR #1266, opened 2019-05-30
Restrict elasticsearch to be less than 7 for searchkick 3.x

Onumis:
Got the error `Faraday::SSLError: SSL_connect returned=1 errno=0 state=SSLv2/v3 read server hello A: unknown protocol` on our apps after upgrading the `elasticsearch` version from 6.3.1 to 7.1.0. (deployed on heroku using bonsai.io fwiw)

This does not happen locally perhaps because SSL is not used in that case.

After restricting `elasticsearch` to 6.x the problem went away. 
I think other people using searchkick 3.x (because they still use Rails 4, for example), might end up with the same problem.

The fix should be made against the 3.x branch, but since there isn't one, can a maintainer please create it?

ps: other dependencies are also open, and perhaps they all should be locked.
```
WARNING:  open-ended dependency on activemodel (>= 4.2) is not recommended
  if activemodel is semantically versioned, use:
    add_runtime_dependency 'activemodel', '~> 4.2'
WARNING:  open-ended dependency on hashie (>= 0) is not recommended
  use a bounded requirement, such as '~> x.y'
WARNING:  open-ended dependency on bundler (>= 0, development) is not recommended
  use a bounded requirement, such as '~> x.y'
WARNING:  open-ended dependency on minitest (>= 0, development) is not recommended
  use a bounded requirement, such as '~> x.y'
WARNING:  open-ended dependency on rake (>= 0, development) is not recommended
  use a bounded requirement, such as '~> x.y'
```

### DIFF
--- a/searchkick.gemspec
+++ b/searchkick.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = ">= 2.2"
 
   spec.add_dependency "activemodel", ">= 4.2"
-  spec.add_dependency "elasticsearch", ">= 5"
+  spec.add_dependency "elasticsearch", ">= 5", "< 7"
   spec.add_dependency "hashie"
 
   spec.add_development_dependency "bundler"


ankane:
Hey @Onumis, thanks for the PR and suggestions 👍 Previous major versions only get fixes for exceptional cases, which is why there isn't a 3.x branch. In this case, since users can fix by updating their Gemfile, I don't think it meets the threshold of an exceptional case (I'm also not sure this affects most users of Searchkick).

```ruby
gem 'elasticsearch', '~> 6'
```

I'm not a fan of bounded requirements in general, as it often unnecessarily prevents developers from upgrading gems.

Onumis:
Sure, that is fine. I just thought that other people might be getting similar errors, since this was a bit hard to understand (linking an SSL error to elasticsearch version). 
But having this issue here is good enough for people to look for and find an answer.
Cheers.

danielpowell4:
@Onumis thanks for submitting! Helped me out for sure.

Think the issue is related to the bonsai addon

Their current versioning is not v7

From their docs (as of 5/31/2019):

| Plan |   Elasticsearch Versions |
| ------------- | ------------- |
| Sandbox |   6.5.4  |
| All other multitenant  |  5.4.3 / 6.0.1 / 6.2.4 / 6.5.4 |


I asked support about a v7 timeline and they said:

> Thanks for reaching out. No timeline yet, but I expect we'll start rolling it out to the Sandbox clusters soon. We may also be able to set it up on a Business subscription right now, but I would need to confirm that.

robsears:
Rob from Bonsai checking in. We've heard about the SSL issue from a few users so far, and it appears to be due to something totally unrelated to Elasticsearch 7. It's a change to the elasticsearch-ruby gem that is being discussed in [this issue](https://github.com/elastic/elasticsearch-ruby/issues/669).

Tl;dr, with the latest version of the gem, you need to explicitly set the port for your Elasticsearch cluster if it's not on 9200. 

So if your cluster has a URL like `https://user:pass@something.com`, the gem no longer infers that it should use port 443, and it attempts 9200. You would need to be explicit and set the URL to `https://user:pass@something.com:443` to make it work.

FWIW Bonsai is planning to start incorporating port 443 into our users' cluster URLs to address this moving forward. We're working on a plan for updating older clusters too.

ankane:
Thanks @robsears 👍

PR #1267, opened 2019-05-30
Elasticsearch Scroll API

kschutt:
## What's up

Reopens #1165

We were running into similar problems to #642 but found that simply increasing the result window was not a viable solution

## What this does

- Adds support for [Elasticsearch Scroll API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html)
- Allows users to pass in the scroll param to initiate a scroll context
- Stores the `scroll_id` on the `Searchkick::Results` object
- Exposes a `scroll` method on `Searchkick::Results` to request the next batch of results

## Notes

- I changed the way the `_scroll_id` is persisted. We always update the `_scroll_id` regardless of which result set is being used. So long as someone doesn't switch back and forth between result sets then it should be okay. I'm not sure how to approach this without requiring the user to provide the `scroll_id` on every request
- Cleaned up the tests a bit
- Added some error handling instead of silently failing


kschutt:
@ankane @ckhall reopening because there might still be a use-case for this. Sorry about the delay

lib/searchkick/query.rb
```diff
@@ -464,6 +469,10 @@ def prepare
         payload[:explain] = options[:explain] if options[:explain]
         payload[:profile] = options[:profile] if options[:profile]
 
+        # scroll
+        raise ArgumentError, "From parameter cannot be used with scroll" if scroll && !payload[:from].nil?
```
ankane: Don't think we need to check for this case - can just let Elasticsearch return an error

lib/searchkick/query.rb
```diff
@@ -464,6 +469,10 @@ def prepare
...
+        payload[:size] = per_page if scroll
```
ankane: It'd be nice to keep this with the rest of the pagination logic.

test/scroll_test.rb
```diff
@@ -0,0 +1,67 @@
...
+    products = products.scroll
+    assert_equal ["Product E", "Product F"], products.map(&:name)
+    assert_equal "product", products.entry_name
+    # scroll exhausted
+    products = products.scroll
+    assert_equal [], products.map(&:name)
+    assert_equal "product", products.entry_name
+  end
+
+  def test_scroll_nil_scroll
```
ankane: Don't think we need this test.

lib/searchkick/results.rb
```diff
@@ -221,6 +221,37 @@ def misspellings?
...
+      else
+        begin
+          params = {
+            scroll: options[:scroll],
+            scroll_id: scroll_id
+          }
+          es_results = Searchkick.client.scroll(params)
+          if scroll_id != es_results['_scroll_id']
+            # update the scroll id so we can keep reusing the same result set
+            @response['_scroll_id'] = es_results['_scroll_id']
```
ankane: I see how this makes the API cleaner, but don't think we should mutate the original response right now. Can revisit at a later time if needed.

kschutt:
@ankane addressed your comments

ankane:
Looks great, thanks @kschutt 👍

ankane:
Also added a block form on master.

```ruby
Product.search("*", scroll: "1m").scroll do |batch|
  # process batch ...
end
```

kschutt:
Sweet! Thanks for cleaning it up and adding the block. 

I wasn’t sure where to add the ability to clear the scroll. If the user doesn’t use the block logic, then they should clear the scroll after their custom logic. Maybe this should be it’s own method on the results object?

ankane:
I'm on the fence about if it's needed (and fine with either direction). I imagine most people will just let the scroll expire, but we could add a method on the results object since the code is already there.

Issue #1262, opened 2019-05-22
Cannot index with custom search_data


yknx4:
Versions: 
ES 7.0
Searchkick 3.1.3
Rails 4.2
Ruby 2.6.1

```
Elasticsearch::Transport::Transport::Errors::BadRequest: [400] {"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"Root mapping definition has unsupported parameters:  [campaign : {_routing={}, dynamic_templates=[{string_template={mapping={ignore_above=30000, type=keyword, fields={analyzed={analyzer=searchkick_index, index=true, type=text}}}, match_mapping_type=string, match=*}}], properties={}}]"}],"type":"mapper_parsing_exception","reason":"Failed to parse mapping [_doc]: Root mapping definition has unsupported parameters:  [campaign : {_routing={}, dynamic_templates=[{string_template={mapping={ignore_above=30000, type=keyword, fields={analyzed={analyzer=searchkick_index, index=true, type=text}}}, match_mapping_type=string, match=*}}], properties={}}]","caused_by":{"type":"mapper_parsing_exception","reason":"Root mapping definition has unsupported parameters:  [campaign : {_routing={}, dynamic_templates=[{string_template={mapping={ignore_above=30000, type=keyword, fields={analyzed={analyzer=searchkick_index, index=true, type=text}}}, match_mapping_type=string, match=*}}], properties={}}]"}},"status":400}
```

this is my code

```ruby
class Campaign < ActiveRecord::Base
  ...
  scope :search_import, -> { includes(:user, :keyword, :target_site, user: [:avatar]) }
  searchkick

  def search_data
    {
      id: id,
      url: url,
      keyword: target,
      country_code: country_code,
      avatar_url: user.avatar.url,
      avatar_thumb_url: user.avatar.thumb.url
    }
  end
  ...
end
```

ankane:
Hey @yknx4, Searchkick 4 is needed for Elasticsearch 7.

yknx4:
@ankane I cannot install version 4 because it depends on ActiveModel 5 thus Rails 5.

You should add to the docs that Rails 4 is not supported in Searchkick 4

PR #1268, opened 2019-05-30
Add Rails >= 5 requirement to Readme

yknx4:
Since Searchkick depends on `ActiveModel > 5` this version won't work on Rails 4 or below, updated readme accordingly.

related to #1262


ankane:
Hey @yknx4, thanks for the PR, but I don't think it makes sense to mention (soon to be fully) unsupported Rails versions in the readme.

PR #1271, opened 2019-06-04
Fixed typo [index_options]

BliZZ4RD:
It looks like typo but without this fix I can't configure custom mappings


ankane:
Thanks @BliZZ4RD, merged in https://github.com/ankane/searchkick/commit/4e0a794044d14ee5ce3a91ffa94043db992a71a9.

Looks like it was introduce last release in https://github.com/ankane/searchkick/commit/5d8cdd0d1a46c89a8359616aa8a183244f4659aa. Will push out a new version shortly.

Issue #1265, opened 2019-05-28
How to call a batch reindex from model


brunoocasali:
Hi, @ankane 

After searching the README and code, I've pretty sure that is there a way to reindex data, inside the `BulkIndexer` but which is not accessible from the public API...

My problem is: 
I have many models in the app with more than 10m rows, but that rows not follow 1, 2, 3, but 1, 49, 87 and goes on... Because we have so many deletions and this breaks the auto increment... 

When I run a reindex with `Question.reindex(async: true, full: true)`, i've the reindex in batches but that batches counts each batch based on a fixed number and the maximum id of the model. This creates a ton of jobs that will do anything like:

```rb
[{"job_class"=>"Searchkick::BulkReindexJob", "job_id"=>"29849376-6c94-496c-adb4-84150958475f", "provider_job_id"=>nil, "queue_name"=>"searchkick", "priority"=>nil, "arguments"=>[{"class_name"=>"Question", "index_name"=>"questions_production_20190528013031810", "batch_id"=>405, "min_id"=>404047, "max_id"=>405046, "_aj_symbol_keys"=>["class_name", "index_name", "batch_id", "min_id", "max_id"]}], "executions"=>0, "locale"=>"pt-BR"}]
```

Looking for a way to deal with it, I've just found the solution in the line: https://github.com/ankane/searchkick/blob/master/lib/searchkick/bulk_indexer.rb#L32 

But I have no ideia how to call this directly from de `Model.reindex` way.

Sorry if i misunderstood the README but I'm pretty sure about that is not accessible, Let me now if could be a new feature or a PR I'm really excited to contribute to this project :D 

Thanks for the awesome gem!

ankane:
Hey @brunoocasali, you should be able to fork the gem and change this line to `elsif false` to do what you want:

https://github.com/ankane/searchkick/blob/a763e06675b0273f6b3e8cef9ce700cb4aaf2895/lib/searchkick/bulk_indexer.rb#L99

Right now, I don't think it's common enough to add an option for.

brunoocasali:
@ankane you remember you I fall into this condition? (without forcing it). Or this piece of code is a "unreachable zone" ?

ankane:
There's not a good way to change the behavior of that conditional without forking it.

BliZZ4RD:
[...]

PR #1277, opened 2019-06-12
Improves BulkIndexer full_reindex_async

BliZZ4RD:
Use .find_in_batches to determinate record_ids that needs to be reindexed #1265


ankane:
Hey @BliZZ4RD, thanks for the PR, but I think the current approach works best for most apps, as it requires a lot less database queries. As I mentioned in #1265, feel free to fork, but it's not something I want to change right now.

PR #1281, opened 2019-06-25
Enable to query on locations not near to specific lon & lat

taman9333:
I ran into a problem which is I have products with their locations "longitude & latitude"based on their vendors.
And I have a page of products which has 2 sections, first section is the products that is near to user's location "which is supported by the gem" so he can buy them directly from the vendors.
the second section is the products that are not near to the user's location so user needs to ship this products with courier to get these products "which is not supported".
I've written unit tests.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -891,6 +891,17 @@ module Searchkick
               case op
               when :within, :bottom_right, :bottom_left
                 # do nothing
+              when :not_near
+                filters << {
+                  bool: {
+                    must_not: {
+                      geo_distance: {
+                        field => location_value(op_value),
+                        distance: value[:within] || "50mi"
+                      }
+                    }
+                  }
+                }  
               when :near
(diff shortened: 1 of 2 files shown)


ankane:
Hey @taman9333, thanks for the PR, including tests 💯 The code looks good, but there's actually a more generalized way to do this already with the `_not` operator.

```ruby
Store.search "*", where: {_not: {location: {near: [37.5, -122.5]}}}
```

taman9333:
Yeah, it works i've read the documentation many times to configure how i could do it & i found nothing.
I was trying (not) but without underscore "_" & it was giving me error.
Thanks for your quick response @ankane

taman9333:
Hi @ankane so could we merge that like another syntax or shall i close the merge request?

ankane:
I don't think it makes sense to add another operator for it. I updated the docs to add `_not`, so hopefully that'll help others. Thanks again for the PR

PR #1285, opened 2019-07-03
Added missing **options argument

PavelBezpalov:
It allow async reindex scoped relations


ankane:
Hey @PavelBezpalov, thanks for the PR 👍 The `async` option was only designed for full reindexing, but think we could make it work for scoped reindexing. Please add a test case for it and make sure the existing tests pass.

ankane:
Cleaning up PRs

PR #1290, opened 2019-07-17
Add match_phrase support to filters

DanielHeath:
This is actually a bit janky (using quotes to signal a phrase match) but it works sort-of OK and does mostly what you'd expect (that is, `where: {body: '"three monkeys"'}` filters to documents containing those words together.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -1024,6 +1024,8 @@ module Searchkick
         end
 
         {regexp: {field => {value: source, flags: "NONE"}}}
+      elsif value.is_a?(String) && value[0] == '"' && value[-1] == '"'
+        {match_phrase: {field => {query: value[1..-2]}}}
       else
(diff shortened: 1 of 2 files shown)


ankane:
Hey @DanielHeath, thanks for the PR 👍 `where` filters operate on untokenized values, so I'm not sure the code works as is. Can you add a test case for it?

DanielHeath:
Added :)

ankane:
Hey @DanielHeath, thanks for updating. `where` is (conceptually) designed to be used with non-analyzed fields, so unfortunately, I don't think this fits into my vision for the project.

DanielHeath:
Fair enough. Is there a sensible alternative for `Find things matching X which also match phrase Y`?

ankane:
There isn't a way to search for multiple terms right now.

ankane:
Cleaning up PRs

Issue #1292, opened 2019-07-29
ES 7.1 Nested Query


BliZZ4RD:
Hi, @ankane.
I have the same problem with the nested query as described in #953. 
Here is my reproduce [script](https://gist.github.com/BliZZ4RD/6e4b3e6b1343cc39464d685a3bff604e) and output

```
Searchkick version: 4.0.2
Elasticsearch version: 7.1.1
Ruby version: 2.6.3
Rails version: 5.2.1

-- create_table(:products)
   -> 0.0132s
-- create_table(:availability_dates)
   -> 0.0003s

{
  "took"=>0, 
  "timed_out"=>false, 
  "_shards"=>{
    "total"=>1, 
    "successful"=>1, 
    "skipped"=>0, 
    "failed"=>0
  }, 
  "hits"=>{
    "total"=>{
      "value"=>0, 
      "relation"=>"eq"
    }, 
    "max_score"=>nil, 
    "hits"=>[]
  }
}

{
  "took"=>0, 
  "timed_out"=>false, 
  "_shards"=>{
    "total"=>1, 
    "successful"=>1, 
    "skipped"=>0, 
    "failed"=>0
  }, 
  "hits"=>{
    "total"=>{
      "value"=>1, 
      "relation"=>"eq"
    }, 
    "max_score"=>1.0, 
    "hits"=>[
      {
        "_index"=>"products_development_20190729220651708", 
        "_type"=>"_doc", 
        "_id"=>"1", 
        "_score"=>1.0
      }
    ]
  }
}
```

LemonAndroid:
It probably won't work directly with the new versions of rails and elasticsearch, but I wrote a tool to more easily debug searchkick via the rails log in the terminal.

Maybe you can get it running: [LogMagic](https://github.com/LemonAndroid/LogMagic)

ankane:
Hey @BliZZ4RD, `where` doesn't work with `nested` mapping types, only nested data. If you remove the custom mapping, the first example in your repro script works.

BliZZ4RD:
@ankane, yep, it works, but instead of searching inside one period it search across all of values

```
Product.create!(
  name: "Test",
  availability_dates: [
    AvailabilityDate.new(start_at: 1.week.ago, finish_at: 1.week.from_now),
    AvailabilityDate.new(start_at: 2.weeks.from_now, finish_at: 3.weeks.from_now)
  ]
)
Product.search_index.refresh
p Product.search(
    "*",
    where: {
[...]

PR #1293, opened 2019-07-30
[Idea] _raw filter support

BliZZ4RD:
Hi @ankane. 
It would be nice to have possibility to modify filter with raw elasticsearch query. With it we can solve some rare cases when Searchkick DSL can't satisfy our needs. It also will prevent us from building whole query body from scratch. Refering to issue #1292 it can be solwed with simple part of query

```
raw_filters = 
  [
    {
      nested: {
        path: 'dates',
        query: {
          bool: {
            must: [
              {
                range: {
                  'dates.start_at' => {
                    lte: '2019-07-01',
                    format: 'yyyy-MM-dd'
                  }
                }
              },
              {
                range: {
                  'dates.finish_at' => {
                    gte: '2019-07-01',
                    format: 'yyyy-MM-dd'
                  }
                }
              }            
            ]
          }
        }
      }
    }
  ]

Product.search('*', where: { warehouse_ids: [1], _raw: raw_filters})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,27 @@ where: {
   _or: [{in_stock: true}, {backordered: true}],
   _and: [{in_stock: true}, {backordered: true}],
   _not: {store_id: 1}           # negate a condition
+  _raw: [
+    {
+      nested: {
+        path: 'dates',
+        query: {
+          bool: {
+            must: [
+              {
+                range: {
+                  'dates.start_at' => {
+                    gte: '2019-07-01',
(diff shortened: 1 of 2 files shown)


ankane:
Hey @BliZZ4RD, thanks for the PR. It's an interesting idea, but I don't think it's super common and there's already a flexible way to do this with advanced search. You can use the block form to avoid writing the entire query from scratch if needed.

ankane:
Cleaning up PRs

Issue #1283, opened 2019-07-01
[Idea] Partial match below - just like misspellings


jfloff:
Would it make sense to have a partial matches `below` keyword, like there is for misspellings? 

`Product.search "fresh honey", operator_below: {operator: "or"; below: 5}`

This would first search `fresh AND honey` and then if there are fewer than 5 results, a 2nd search is performed with opertor `OR`

ankane:
Hey @jfloff, makes sense to me if you want to put together a PR. It should probably trigger after `misspellings` below. I also think we can reuse the existing `operator` option instead of creating a new `operator_below`.

jfloff:
I will take a jab at the PR :)

Regarding the option, how should we approach the `default` vs the `below` operator then? How about this? 
`Product.search "fresh honey", operator: { default: "AND", below: { "OR" : 5 " } }`

ankane:
I'm thinking we can just do `operator: {below: 5}`. It should always go from `AND` to `OR`.

jfloff:
Hi there, Just a heads up that at the moment I'm a bit swamped with work in other areas. Just feel free - if you don't want to keep the issue open - to close the issue, and then when I'm done with the new feature I will just submit a PR.

ankane:
👍

PR #1304, opened 2019-08-27
Partial matches below theshold

jfloff:
Idea discussed #1283 

I had to make a few more changes than I was expecting initially due to the behavior when we have both `misspellings: { below: X }`  and `operator: { below: X }` - when we execute one, both, in which order, etc. The tests I added should help understand some of the cases I had to work around.

Like you mentioned in #1283 the misspellings retry gets precedence over the operator retry. I added a `retry_order` that at this moment is hard coded, but I can see someone wanting to change this order in the near future, hence I tried making this PR ready for that.

Hope it fits what we discussed, feel free to code review and suggest changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,15 @@ To change this, use:
 Product.search "fresh honey", operator: "or" # fresh OR honey
 ```
 
+Or if you prefer, Searchkick can first perform a search where results match all the words, and if there are too few results, perform another with partial matches enabled.
+
+```ruby
+Product.search "fresh honey", operator: {below: 5}
+```
+
+If there are fewer than 5 results, a 2nd search is performed with the `or` operator. The result of this query is returned.
+
+
 By default, results must match the entire word - `back` will not match `backpack`. You can change this behavior with:
 
 ```ruby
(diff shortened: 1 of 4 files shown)


ankane:
Hey @jfloff, thanks for the PR 👍 Will check it out when I have some time.

jfloff:
Hey @ankane have you looked into this?

ankane:
Hey @jfloff, really sorry for the long delay. I've reviewed this a few separate times, but am still worried about the complexity the feature in general adds (should have caught this when we originally discussed). For more advanced cases, I think it might be better for users to do:

```ruby
products = Product.search(term, **options, misspellings: false)
products = Product.search(term, **options, misspellings: true) if products.count < 5
products = Product.search(term, **options, operator: "or") if products.count < 5
```

This would allow for simpler logic in Searchkick and I think make it more clear what's going on in application code. One downside is the `track` option isn't suited for this, so users would probably want to record searches manually. Let me know your thoughts.

jfloff:
I guess that works. Thanks for the feedback @ankane

PR #1319, opened 2019-10-01
Added search_after parameter handling

sineed:
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#request-body-search-search-after


ankane:
Hey @sineed, thanks for the PR, including a test 💯 The code looks good. This feature can also be accomplished with:

```ruby
Product.search body_options: {search_after: ...}, ...
```

Since I don't think it's a common option and there's a more general way to do it already, I don't think it makes sense to merge right now.

sineed:
Thank you for the explanation, I didn't know that. Sure, PR doesn't make sense

PR #1331, opened 2019-10-25
Buld reindexing with custom batches

ombr:
Hi,

I need to change the batching logic of the bulk indexer (custom sorting with priorities)

I am looking todo something like this:

```
  def self.bulk_reindex(index_name = nil)
    bulk_indexer = Searchkick::BulkIndexer.new(SearchKick::Index.new(index_name))

    loop_by_ids do |ids|
      indexer.bulk_reindex_job(Product, SecureRandom.base64, record_ids: ids)
      Rails.logger.info "Indexing: #{Product.where(id: ids.last).pluck(:created_at).first} #{batch_id}"
    end
    indexer.wait_for_completion do |status|
      Rails.logger.info "Batches left: #{status[:batches_left]}"
    end
  end
```

I did the following change after reading the code (moving some responsabilities).

Tests are passing.

Do you think it is a good idea ? How can I improve it ?

### DIFF
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -169,20 +169,18 @@ module Searchkick
   end
 
   def self.reindex_status(index_name)
-    raise Searchkick::Error, 'Redis not configured' unless redis
-
-    Searchkick::Index.new(index_name).bulk_indexer.status
+    Searchkick::Index.new(index_name).reindex_status
   end
 
   def self.with_redis
-    if redis
-      if redis.respond_to?(:with)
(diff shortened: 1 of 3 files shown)


ankane:
Hey @ombr, thanks for the PR 👍 It looks like this change only affects the internal APIs, which shouldn't really be used (as they could break at any time), so I don't think it makes sense to merge. I've done some prototyping to support more complex indexing patterns, but this will likely be a Pro feature if released. That being said, you can always run your own fork with the patch above.

PR #1350, opened 2019-12-11
Allow Custom Load Option to be Set in Model

berniechiu:
### Description

At some point, ES have all the data we want and we don't have needs to query from DB again. But I don't want to explicitly set the `load: false` options everytime when we search. I wonder if this is a good place to put the custom option into 😌


ankane:
Hey @berniechiu, thanks for the PR! This is a nice feature. However, I think users will expect most search options to work if any of them do, so I don't think it can be added alone.

I've considered adding a `search_options` key (similar to `default_scope` for ActiveRecord) in the past, but I'm worried about the complexity it adds. For options like `where` and `misspellings`, we'd need to decide whether or not to merge the default options (probably more intuitive but much more complex) and how unscoping works. Also, I worry it'll lead to more confusion with users (default scopes in ActiveRecord seem to cause issues that people don't expect / realize that's the cause).

Based on the above, I think the best way to accomplish this is to create a custom search method in the model that applies any user defaults.

```ruby
  def my_search(term)
    search(term, load: false)
  end
end
```

berniechiu:
cool, sounds fair enough

PR #1367, opened 2020-01-30
Fix argument error when passing `refresh` option to `reindex`

milani:
`reindex_scope` function that is called in the else block of `reindex` raises an argument error exception when `refresh` option is passed on. This commit makes sure it is removed from the options.

### DIFF
--- a/lib/searchkick/index.rb
+++ b/lib/searchkick/index.rb
@@ -184,6 +184,7 @@ module Searchkick
 
     def reindex(relation, method_name, scoped:, full: false, scope: nil, **options)
       refresh = options.fetch(:refresh, !scoped)
+      options.delete(:refresh)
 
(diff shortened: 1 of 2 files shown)


ankane:
Hey @milani, thanks for the PR. Can give an example of code that's causing the error (or add a test case)? Also, you'll want to delete the key in a separate line, as this changes the current behavior with `refresh: false`.

milani:
@ankane I added a test that would fail without the fix. I also do the delete in a separate line.

ankane:
Thanks for updating. Can the test case be simplified to `Product.reindex(refresh: true)`?

milani:
@ankane done.

ankane:
Awesome, thanks @milani!

Issue #1353, opened 2019-12-16
[Idea] Support "refresh: wait_for" option when reindexing


mattvague:
Hey there, in the app I'm working on we have the requirement that if we query elastic search immediately after creating a new record, that this record shows up. Because of the 1s delay between index refreshes this is unreliable as-is.

I've explored several different options and it seems like the best one is to use ES's [refresh: wait_for](https://www.elastic.co/guide/en/elasticsearch/reference/master/docs-refresh.html#docs-refresh) option during index updates. This will make it so that our API will not respond after a request until elastic search has finished refreshing.

Is there any chance of adding a new `callbacks: :wait_for` option or something? Failing that, would you accept a pull request to allow a bit more configurability in Searchkick to allow us to implement this ourselves?

Thanks!

ankane:
Hey @mattvague, a new callback mode makes sense to me. Happy to review a PR.

mattvague:
I was exploring this, and I don't think it will be too difficult besides figuring out how to pass this new "refresh" option through all the way through to `Searchkick::Indexer` so we can do `Searchkick.client.bulk(body: items, refresh: refresh_mode)`. 

What would you advise here? Store the option on `Searchkick::Index` and then add option arguments to `Searchkick::Indexer#queue` and `Searchkick::Indexer#perform`?

ankane:
I don't fully follow the description/question. I think this one will be easier to discuss when there's code.

mattvague:
> I don't fully follow the description/question. I think this one will be easier to discuss when there's code.

Hah, wasn't my finest piece of writing. For sure, will throw up a PR for you to look at

ankane:
Cleaning up issues, but still happy to review a PR.

TikiTDO:
@ankane I'm looking at throwing together a PR for this issue, but I don't see a way to add this as a callback mode without adding a whole lot of confusion around testing.

Would you be open to adding a new top-level config option, something similar to `Searchkick.client_options`, but more like `Searchkick.refresh_mode`?

PR #1380, opened 2020-02-17
[1353] Refresh mode option

TikiTDO:
Example of potential solution to #1353

### DIFF
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -35,7 +35,7 @@ module Searchkick
   class ImportError < Error; end
 
   class << self
-    attr_accessor :search_method_name, :wordnet_path, :timeout, :models, :client_options, :redis, :index_prefix, :index_suffix, :queue_name, :model_options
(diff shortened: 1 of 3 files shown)


TikiTDO:
Here's what I was thinking for this feature. Let me know if you have a better idea of where to put the param for this mode.

TikiTDO:
Could even make it more generic, like `bulk_import_options` or even just `import_options` that you could merge into the request.

ankane:
Hey @TikiTDO, thanks for this PR as well! Based on the discussion in #1353, my initial thoughts are this should be a new reindex `mode` rather than a global option.

TikiTDO:
Unfortunately I've now moved on to other parts of our system after we monkey-patched this functionality in an initializer. Since then, I haven't worked with this library in over a month which means it's all but faded from memory at this point.

I can revisit this if I return back to indexing at some point, but my current focus is an entirely different system in a different language, so  I'm not sure when I fill have the time to figure out how and where to hook into the index mode, and also write some good tests for that.

ankane:
No worries, I appreciate you taking a shot at it

PR #1381, opened 2020-02-17
Clarify how callbacks work in the README

TikiTDO:


ankane:
Hey @TikiTDO, thanks for the PR! Can you explain this a bit more? There's not anything in that section about changing the default behavior, so I'm not sure it makes sense to have instructions for how to restore the default behavior.

TikiTDO:
@ankane This PR was the result of hours of searching through the code-base after someone I work with had an issue running tests while using searckick. For most of out tests we disable callbacks, because we usually don't need to test indexing behavior, as so:

```
Searchkick.callbacks(false) do
   test.call
end
```

however, for some tests we actually wanted to restore the default behavior for at least parts of the test, and figuring how to do that was the challenge.

There is nowhere in the docs that explains that default behavior is `Searchkick.callbacks(nil)`, so we used `Searchkick.callbacks(true)`, `Searchkick.callbacks(:bulk)`, and variations thereof. All of these resulted in really weird and hard to debug edge-cases that would always differ from what happened in production, which is what let me to a few hours of digging through the code to figure out how all this worked.

The point of this PR is to document the fact that there is another option that may be passed into the `Searchkick.callbacks` function; I find it important to document all such valid options, particularly in code that is released publicly. This bit of documentation would have saved us a combined 10-12 programmer hours, so I figured I'd send this in to save other people the pain in the future.

ankane:
It sounds like this is related to testing, so think it makes more sense to update that section if needed. Can you explain your setup a bit more (test gem, etc) and how if differs from the examples in the readme? https://github.com/ankane/searchkick#testing

TikiTDO:
We use rspec. 

The testing example shows:

```ruby
RSpec.configure do |config|
  config.before(:suite) do
    # reindex models
    Product.reindex

    # and disable callbacks
    Searchkick.disable_callbacks
  end

  config.around(:each, search: true) do |example|
    Searchkick.callbacks(true) do
      example.run
    end
  end
end
```

Based on that, some of our tests were wrapped with:

```ruby
    Searchkick.callbacks(false) do
      example.run
    end
```

And we had blocks where we wanted to simulate production behavior, by doing:

```ruby
    Searchkick.callbacks(nil) do
       # Some stuff to test
    end
```

Since the callbacks section of the docs did not mention that `nil` had a special meaning, we figured `true` was synonymous with `on`, and instead decided that this was some weird rspec bug. Basically, I got tasked with searching through rspec, and a bunch of other code to figure out why we were getting different behavior from production when we did:

```ruby
    Searchkick.callbacks(true) do
       # Some stuff to test
    end
```

ankane:
Thanks @TikiTDO. It sounds like the example should be updated from `Searchkick.callbacks(true)` to `Searchkick.callbacks(nil)` to restore the original behavior.

ankane:
Updated the example as mentioned above, and will shorten it to `Searchkick.callbacks` (no `nil`) in the next release to make it slightly cleaner. Thanks for pointing out it `nil` is a better option.

PR #1383, opened 2020-02-20
Boundary character options in highlighting 

naveed-ahmad:
If content has html tags, highlighting usually break the format because of default [boundary chars](https://www.elastic.co/guide/en/elasticsearch/reference/7.5/search-request-body.html#highlighting-settings) options`.,!? \t\n.`.

Example: 
```
  store_names ["Two <strong>Door Cinema Club Some Other Words And Much More Doors Cinema</strong>Club"]

    Product.search("cinema", highlight: true).first.search_highlights[:name]
   => 
<strong>Door <em>Cinema</em> Club Some Other
```

Note: no closing `</strong>` tag in highlight result. And this will break whole document sometime.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -767,6 +767,11 @@ module Searchkick
         if (fragment_size = options[:highlight][:fragment_size])
           payload[:highlight][:fragment_size] = fragment_size
         end
+
+        if (boundary_chars = options[:highlight][:boundary_chars])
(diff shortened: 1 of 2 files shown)


ankane:
Hey @naveed-ahmad, thanks for the PR! We should probably allow any of the highlighting options to be set, so they don't need to be added individually. Also, it looks like the tests are failing.

ankane:
Cleaning up stale PRs

Issue #1396, opened 2020-03-18
Elasticsearch 7.5 support?


NicoSa:
Hi,

we are trying to upgrade our project to Ruby 2.7.0 however we are finding problems with Searchkick when we do that. We think the reason for this might be that only Elasticsearch-Ruby 7.5 supports Ruby 2.7.0 but currently Searchkick works with 7.4.

Could we be seeing issues there? When is Searchkick going to support Elasticsarch-Ruby 7.5?

Thanks!

ankane:
Hey @NicoSa, Searchkick works fine with Elasticsearch 7.5. https://travis-ci.org/github/ankane/searchkick/builds/663745404.

If you're running into a specific error, please use the gist in the Contributing Guide to reproduce.

NicoSa:
Hey @ankane, I think the problem is Ruby 2.7.0 though. Are you able to verify that Searchkick works with Elasticsearch on Ruby 2.7.0 ?

ankane:
It works on Ruby 2.7 as well.

NicoSa:
Ok, must be some other issue then. I'll get back to you if I can trace it down to Searchkick. So far so good, thank you @ankane !

PR #1397, opened 2020-03-18
test ruby 2.7 as well

ChristianHeimke:
according to #1396 I've added ruby 2.7 to the travis.ci configuration to test it too and it seems to be working  :+1:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 dist: bionic
 language: ruby
-rvm: 2.6
+rvm:
+  - 2.6
+  - 2.7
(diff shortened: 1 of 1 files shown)


ankane:
Hey @ChristianHeimke, thanks for the PR! To keep the test suite fast and simple, I like to only test against a single Ruby version (which hasn't been an issue in the past), and have been waiting for the next Rails release to remove deprecation warnings before bumping to 2.7.

ankane:
Just pushed an approach to get wide Ruby coverage while keeping the test suite fast.

Issue #1411, opened 2020-04-23
Missing Aggregation


kelevro:
Implement [Missing Aggregation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-missing-aggregation.html#search-aggregations-bucket-missing-aggregation)
it will be great if we can specify `missing` option for aggregation field to get docs count with empty value
```ruby
Product.search "wingtips", aggs: {color: {missing: true}}
```

PR #1412, opened 2020-04-23
add missing option

kelevro:
Implementation [Missing Aggregation](https://github.com/ankane/searchkick/issues/1411)

### DIFF
--- a/README.md
+++ b/README.md
@@ -764,6 +764,12 @@ Order
 Product.search "wingtips", aggs: {color: {order: {"_key" => "asc"}}} # alphabetically
 ```
 
+By default, documents with `null` value are skipped, to include use [missing](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-missing-aggregation.html) option
+
+```ruby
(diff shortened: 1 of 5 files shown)


ankane:
Hey @kelevro, thanks for the PR! `body_options` is preferred for different aggregation types so they don't need to be added individually.

kelevro:
@ankane thanks for you feedback! it's great that we can use both `aggs` and `body_options`
```ruby
Product.search(
  "wingtips", 
  aggs: [:color], 
  body_options: {missing_color: {missing: {field: 'color'}}}
)
```
maybe we should document this option
but I think it will be great if we be able to use aggregation conditions in both cases
```ruby
Product.search(
  "wingtips", 
  aggs: {color: {where: { store_id: 1 }}}, 
  body_options: {
    aggs: {
     missing_color: {
       filter: {bool: {must: term: {store_id: 1}}},
       aggs: {missing_color: {missing: {field: 'color'}}}
    } 
   }
 }
)
```
```ruby
Product.search(
  "wingtips", 
  aggs: {color: {where: { store_id: 1 }, missing: true}}
)
```
what do you think?

PR #1440, opened 2020-09-12
Skip optional regex operators when using `:like` syntax

tbrisker:
Similar to the fix in 6664a9ea3fc8bf70210937bfd619666672bde220, this
also sets the optional Lucene regex flags to "NONE" when using :like
syntax for searching. Otherwise, searches for strings including any of
the optional operators (# @ & < > ~) causes issues since they are not
escaped.


ankane:
Looks great, thanks @tbrisker!

tbrisker:
Thank you so much for the quick merge @ankane, I appreciate your hard work maintaining this gem! Do you happen to have any idea approximately when you plan to do a new release with this fix? just so we know if we should invest in working around it for now or rather wait for the proper fix.

ankane:
I'm not sure when the next release will be, but you can point your Gemfile to GitHub if needed.

PR #1442, opened 2020-10-05
[1353] Refresh mode option try 2

TikiTDO:
See discussion in #1380, in reference to #1353 

I wasn't sure what you meant by "this should be a new reindex mode rather than a global option", so I implemented a handler similar to how you handle callbacks option. If this is not what you meant, could you provide a snippet of the syntax you would find acceptable.

Edit: TODO - lib/chewy/type/import/bulk_request.rb:38

### DIFF
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -162,6 +162,20 @@ module Searchkick
     end
   end
 
+  def self.wait_for_refresh(value = true)
+    if block_given?
+      previous_value = should_wait_for_refresh
+      begin
+        self.should_wait_for_refresh = value
+        yield
+      ensure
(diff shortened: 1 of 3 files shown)


TikiTDO:
@ankane I'll be working on stuff related to this gem for a couple of weeks. If you can offer some feedback/snippets showing what you'd like this to look like then I'll be happy to get this PR up to spec in the next little while. Otherwise I worry this might end up like last time when I had forgotten everything I had previously researched when you looked at my PR after a few months.

TikiTDO:
I originally needed this during tests, in order to get valid results after a batch import. It turns out that you can get similar results by doing `Model.search_index.refresh` in the test after doing a batch import, which satisfies my needs.

PR #1450, opened 2020-11-07
Use `.search_import` in ReindexV2Job

trevororeilly:
Avoids N+1 queries when loading grandchild associations (e.g.
Product -> Store -> Address).

### DIFF
--- a/lib/searchkick/reindex_v2_job.rb
+++ b/lib/searchkick/reindex_v2_job.rb
@@ -11,19 +11,7 @@ module Searchkick
 
     def perform(klass, id, method_name = nil, routing: nil)
       model = klass.constantize
-      record =
-        begin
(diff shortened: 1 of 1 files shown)


ankane:
Hey @trevororeilly, thanks for the PR! This makes sense, but would like to hold off until Searchkick 4 to reduce the chance of causing issues. I think we can probably replace `unscoped` with `search_import` in that upgrade.

ankane:
Hey @trevororeilly, just released Searchkick 5 (meant to say that above, not 4), which adds the scope (there were significant changes to the reindexing logic, so it ended up in a different class). Thanks again for the PR!

trevororeilly:
@ankane Great! Appreciate the follow up.

PR #1458, opened 2021-01-27
Fix sentence in Queuing section of README

artinboghosian:
Just found a sentence that was incorrect, used indexed instead of indexing.


ankane:
Thanks @artinboghosian

PR #1465, opened 2021-02-18
Multiple clients

sharshenov:
Currently, the Searchkick can work with a single ES cluster only. Having ability to use separate ES clusters per AR model could be helpful for gradual migration between ES clusters.

In order to use custom client:
1. Define client in the initializer
```ruby
Searchkick.add_client :custom_client, url: 'https://user:pass@example.com/'
```

2. Configure model to use custom client
  ```ruby
class MyModel < ApplicationRecord
    searchkick client_name: :custom_client
end
```

Same approach (but different implementation for outdated version of Searchkick) [helped Veeqo to gradually migrate model-by-model between 5.6 and 6.8 ES clusters](https://devs.veeqo.com/a-veeqo-journey-to-become-a-better-elasticsearch-user/).

It is a draft PR. Not sure if the feature is desired for upstream. Should I proceed polishing it and adding more tests and docs?

### DIFF
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -40,9 +40,13 @@ module Searchkick
   class ImportError < Error; end
 
   class << self
+    extend Forwardable
+
     attr_accessor :search_method_name, :wordnet_path, :timeout, :models, :client_options, :redis, :index_prefix, :index_suffix, :queue_name, :model_options
     attr_writer :client, :env, :search_timeout
     attr_reader :aws_credentials
+
+    def_delegators :client, :server_version, :server_below?
   end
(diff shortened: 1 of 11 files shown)


ankane:
Hey @sharshenov, thanks for sharing. That article is really good!

In terms of upgrading, most users should be able to do rolling upgrades. Multiple clients is a neat idea, but I don't think it's worth the extra complexity to support it.

sharshenov:
I understand.
Anyway, thank you for the Searchkick and other projects! :star:

sdhull:
FWIW this would be really cool to have. I'm quite certain we'd use it at my company

PR #1466, opened 2021-03-04
Added support for special chars to include punctuations and symbols

gist-support:
Thanks for contributing. You’re awesome! A few things to keep in mind:

- Keep changes to a minimum
- Follow the existing style
- Add one or more tests if possible

Finally, replace all this with a description of the changes.


ankane:
Hey @gist-support, thanks for the PR. Can you add a description and a test case? Also, it looks like this breaks existing tests.

gist-support:
Pull request updated incorrectly

Issue #1469, opened 2021-03-10
Enabling emoji does not generate expected query for multi-word emoji symbols


rbviz:
quote from https://github.com/ankane/searchkick#emoji

> Search 🍨🍰 and get ice cream cake!

actually results in `ice_cream cake`. 

```ruby
$> ModelName.search("🍨🍰", emoji: true, execute: false).params[:body][:query]
=> {:bool=>{:should=>[{:dis_max=>{:queries=>[{:multi_match=>{:query=>"ice_cream  cake", :boost=>10, :operator=>"and", :analyzer=>"searchkick_search", :fields=>["*.analyzed"], :type=>"best_fields"}}, {:multi_match=>{:query=>"ice_cream  cake", :boost=>10, :operator=>"and", :analyzer=>"searchkick_search2", :fields=>["*.analyzed"], :type=>"best_fields"}}, {:multi_match=>{:query=>"ice_cream  cake", :boost=>1, :operator=>"and", :analyzer=>"searchkick_search", :fuzziness=>1, :prefix_length=>0, :max_expansions=>3, :fuzzy_transpositions=>true, :fields=>["*.analyzed"], :type=>"best_fields"}}, {:multi_match=>{:query=>"ice_cream  cake", :boost=>1, :operator=>"and", :analyzer=>"searchkick_search2", :fuzziness=>1, :prefix_length=>0, :max_expansions=>3, :fuzzy_transpositions=>true, :fields=>["*.analyzed"], :type=>"best_fields"}}]}}]}}
```

Searchkick versions (4.4.2, 4.4.3)
    gemoji (3.0.1)
    gemoji-parser (1.3.1)

PR #1470, opened 2021-03-10
Emoji matching without misspelling correction

rbviz:
Splits emoji token names into words and enables emoji support without relying on misspelling correction

Fixes: https://github.com/ankane/searchkick/issues/1469


ankane:
Looks great, thanks @rbviz!

Issue #1491, opened 2021-06-10
`model_name` throws an error when Searchkick::Results contains multiple models


nathunsmitty:
**Describe the bug**
When searching across [multiple models](https://github.com/ankane/searchkick#multiple-models), the returned Searchkick::Results object will throw the following error:

```
*** NoMethodError Exception: undefined method `model_name' for nil:NilClass
```

This is because `klass` is not defined on Searchkick::Results with multiple models. Since `model_name` [depends on `klass`](https://github.com/ankane/searchkick/blob/df1364c890113259f26116389896e941eebbd81d/lib/searchkick/results.rb#L73), an error is raised here when `klass` is not defined.

I first noticed this issue with Kaminari's `page_entries_info`, which relies on `model_name` in order to work.

**To reproduce**
Use this code to reproduce when possible:

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"

  gem "activerecord", require: "active_record"
  gem "activejob", require: "active_job"
  gem "sqlite3"
  gem "searchkick", git: "https://github.com/ankane/searchkick.git"
end

puts "Searchkick version: #{Searchkick::VERSION}"
puts "Elasticsearch version: #{Searchkick.server_version}"

ActiveRecord::Base.establish_connection adapter: "sqlite3", database: ":memory:"
ActiveJob::Base.queue_adapter = :inline

ActiveRecord::Schema.define do
  create_table :products do |t|
    t.string :name
  end

  create_table :categories do |t|
    t.string :name
  end
end

class Product < ActiveRecord::Base
  searchkick
end

class Category < ActiveRecord::Base
  searchkick
end

Product.reindex
Category.reindex
Product.create!(name: "Apple Sauce")
Category.create!(name: "Apple Products")
Product.search_index.refresh
Category.search_index.refresh
results = Searchkick.search "apple", models: [Product, Category]
results.model_name
```

**Additional context**
My suggestion for a fix would be to default to "Result" if `klass` is undefined in Searchkick::Results. Happy to make a PR for this if you'd like. Open to other alternative fixes as well!

This issue is somewhat related to #139.

ankane:
Hey @nathunsmitty, thanks for reporting! `Result` makes sense to me. Happy to review a PR.

nathunsmitty:
Awesome! PR up in #1492

ankane:
PR merged 🎉

PR #1492, opened 2021-06-11
Use `Searchkick::Results` as class for `model_name` when searching across multiple models

nathunsmitty:
As noted by #1491, `Searchkick::Results.model_name` returns an error if searching across multiple models. This PR adds a conditional to ensure that in this case, `model_name` falls back to just returning an `ActiveModel::Name` on itself. In practice, this means that `results.model_name.human` will return `Results` instead of an error.

One thing I wasn't totally sure about was whether to keep the model name as plural ("Results") or singularize it to ("Result"). Ultimately, since the class is named "Results" I decided to stay with the plural but note that this does slightly differ from the singe model search case which would return "Product". I don't care too strongly either way though and am happy to change it – I'm mainly interested in this method because it's used internally by Kaminari, but Kaminari's pretty good about calling singularize/pluralize on the model name depending on the number of results so the model name's pluralization doesn't matter in this case.

### DIFF
--- a/lib/searchkick/results.rb
+++ b/lib/searchkick/results.rb
@@ -1,5 +1,5 @@
 require "forwardable"
-require "active_support/core_ext" # Required by active_model/naming
+require "active_support/core_ext/module/delegation" # Required by active_model/naming
 require "active_model/naming"
 
 module Searchkick
--- a/lib/searchkick/results.rb
+++ b/lib/searchkick/results.rb
@@ -74,7 +74,7 @@ module Searchkick
 
     def model_name
       if klass.nil?
-        ActiveModel::Name.new(self.class)
+        ActiveModel::Name.new(self.class, nil, 'Result')
       else
         klass.model_name
       end
--- a/lib/searchkick/results.rb
+++ b/lib/searchkick/results.rb
@@ -1,4 +1,6 @@
 require "forwardable"
+require "active_support/core_ext" # Required by active_model/naming
+require "active_model/naming"
 
 module Searchkick
   class Results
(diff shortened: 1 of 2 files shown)


ankane:
Hey @nathunsmitty, thanks for the PR! A few thoughts:

1. I think it should probably be singular (`Result`) like a model would be
2. I'd prefer not to add any additional `require` statements

nathunsmitty:
Hey @ankane! Sorry for my delay – I started a new job last week, so things have been a little busy. Re: your comments,

1. I updated `Results` to have a singular class name.
2. Unfortunately, [Rails' `model_name`](https://api.rubyonrails.org/classes/ActiveModel/Naming.html#method-i-model_name) specifically returns an `ActiveModel::Name` object. I'm not sure how we can get around importing the definition. It is regrettable that `active_support/core_ext/module/delegation` isn't `require`d by by `active_model/naming` even though `active_model/naming` uses the `delegate` method – I think that might be a bug with Rails. I can file an issue in Rails about that, but I think that `active_model/naming` is a hard dependency for this fix.

Of course, I'm all pretty new to Ruby and Rails, so let me know your thoughts! It's very possible I'm missing something obvious.

nathunsmitty:
Hey @ankane! Just wanted to check in if you had any thoughts on this PR as updated. :)

ankane:
Hey @nathunsmitty, sorry for the delay. Looks good overall. Can you remove the `require` statements? `ActiveModel::Naming` should already be loaded by the time the method is called (and if it's not in certain cases, we can handle it later).

nathunsmitty:
Done! Confirmed that `ActiveModel::Name` is loaded without the statements for both the tests and in my rails app 😄

ankane:
Looks great, thanks @nathunsmitty!

Issue #1495, opened 2021-06-25
Use sorted set for ReindexQueue instead of list


willcosgrove:
Would you be interested in a PR to change the Redis data type being used to store the ReindexQueue? It's currently stored in a list, which achieves the need to make sure every item gets processed: push to one end of the list, pop from the other end. But the current implementation is very chatty with Redis. It calls `rpop` in a loop.

https://github.com/ankane/searchkick/blob/6ea0c2fd27dff3d07c06f29356e03fc864680bf3/lib/searchkick/reindex_queue.rb#L16-L22

(It also should probably hold that Redis connection it gets in the `with_redis` block for the entirety of the loop instead of requesting a new connection on each iteration, but that can be addressed separately.)

At first, I thought this would be a good candidate to use the new `count` parameter added to `rpop` in Redis 6.2. But doing that would still potentially require many `rpop` calls because of the deduplication happening by pulling the records off into a set.

Since it doesn't benefit us to store multiples of the same ID to reindex, we can push that set behavior to Redis while maintaining our queue-like access by using a sorted set.

The new implementation would essentially just be:

```rb
def push(record_id)
  Searchkick.with_redis { |r| r.zadd(redis_key, Time.now.to_f, record_id, nx: true) }
end

def reserve(limit: 1000)
  Searchkick.with_redis { |r| r.zpopmin(redis_key, limit) }
    .map { |(id, _score)| id }
end

def clear
  Searchkick.with_redis { |r| r.del(redis_key) }
end

def length
  Searchkick.with_redis { |r| r.zcard(redis_key) }
end
```

And for the work with the reindex refactor, the `push` method can handle multiple records like this:

```rb
def push(*record_ids)
  now = Time.now.to_f
  record_ids.map! { |id| [now, id] }
  Searchkick.with_redis { |r| r.zadd(redis_key, record_ids, nx: true) }
end
```

The reason why you may not want to make this change is that there is extra work involved in figuring out how to handle the upgrade process. I can think of two reasonable options:

1. **Make the upgrade opt-in**
   
   The user would have to now configure what ReindexQueue implementation to use, with the default needing to be the current List implementation.
2. **Make the upgrade mandatory but transparent**
   
   Add additional migration handling code that would detect if the backing key was not a sorted set, and if not would run some code to move the records into a sorted set before continuing.

   In the past when I've done things like this, I will usually include a Redis lua script that is capable of performing a data migration atomically.

[...]

PR #1498, opened 2021-07-06
Performance improvement for ReindexQueue#reserve when using Redis >=6.2

willcosgrove:
This incorporates the improvements discussed in #1495.

If the Redis server is >= 6.2, then we will use the new RPOP count argument to reduce the calls to Redis down to 1. Otherwise, we still need to talk to Redis `limit` number of times. One small optimization has been made in this case to only reserve a redis client one time, rather than `limit` number of times.

I went back and forth on another potential optimization but decided to leave it out. If we assume that the reindex queue is normally larger than the reserve limit, then it is faster to pipeline the `rpop` calls and compact the result, than it is to `rpop` one at a time and check for nil values.

```rb
def reserve(limit: 1000)
  if supports_rpop_with_count?
    # ...
  else
    Searchkick.with_redis do |r|
      r.pipelined { limit.times { r.rpop(redis_key) } }.compact
    end
  end
end
```

But this has a performance degradation in the case that the reindex queue is normally empty or at least much less than the reserve limit. It will end up sending `limit` requests to Redis regardless of the list size.

In an effort to not negatively change the performance characteristics for any case, I've left it as is. If users want performance improvements here, the best option is to upgrade Redis to take advantage of the `RPOP` count.

### DIFF
--- a/lib/searchkick/reindex_queue.rb
+++ b/lib/searchkick/reindex_queue.rb
@@ -14,11 +14,17 @@ module Searchkick
 
     # TODO use reliable queuing
     def reserve(limit: 1000)
-      record_ids = Set.new
-      while record_ids.size < limit && (record_id = Searchkick.with_redis { |r| r.rpop(redis_key) })
-        record_ids << record_id
+      if supports_rpop_with_count?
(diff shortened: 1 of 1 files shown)


ankane:
Looks great, thanks @willcosgrove!

PR #1500, opened 2021-07-07
Add the max_id option to resume

ogidow:
We are creating an id field so we can know the correct max id.
In that case, it would be nice to be able to specify the max id externally so that we can resume the reindex correctly.

### DIFF
--- a/lib/searchkick/bulk_indexer.rb
+++ b/lib/searchkick/bulk_indexer.rb
@@ -24,7 +24,13 @@ module Searchkick
           # to get the max _id without scripting since it's a string
 
           # TODO use primary key and prefix with table name
-          relation = relation.where("id > ?", index.total_docs)
+
+          max_id = index.total_docs
+          if resume.is_a?(Hash) && resume[:max_id]
+            max_id = resume[:max_id]
+          end
+
+          relation = relation.where("id > ?", max_id)
         end
 
         relation = relation.select("id").except(:includes, :preload) if async


ankane:
Hey @ogidow, thanks for the PR. Can you share your reindex logic of how this is used?

ogidow:
@ankane 
Sorry for the late reply

We have added various conditions for indexing in the should_index? method
so there will be a difference between the actual id that should be resumed and index.total_docs

```
id should_index?
1 true
2 true
3 false
4 false
5 false
6 true
7 true
8 true
```

For example, in the above example, only the data for which should_index? is true will be indexed, so the actual id of the data to be indexed will be 1,2,6,7.
If reindex fails with id 7, the next id to be reindexed is 7, but total_docs is 3, so reindex will be restarted from id 3.
This leads to a reindex overhead.
For small datasets, I think this problem can be ignored, but
For large data sets, we expect this overhead to be quite large.

Therefore, I would like to add the max_id option to notify the actual id that reindex should be restarted from the outside.

Also, since we are storing the DB id data in the Elasticsearch  field, we expect to use the max_id option as follows

```rb
SearchkickModel.reindex
# reindex failed due to some problem

max_id = SearchkickModel.elasticsearch_max_id # Method to get the maximum id stored in Elasticsearch
SearchkickModel.reindex(resume: {max_id: max_id})
```

ankane:
Hey @ogidow, can you share the `elasticsearch_max_id` method?

PR #1502, opened 2021-07-14
Replace uniq with distinct in the README

raulr:
Just found that `uniq` is used in several code samples in the README, but `Relation#uniq` [was deprecated in Rails 5.0](https://guides.rubyonrails.org/5_0_release_notes.html#active-record-deprecations) and removed in Rails 5.1. So using `Relation#distinct` should be the way to go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -649,7 +649,7 @@ class Product < ApplicationRecord
   def search_data
     {
       name: name,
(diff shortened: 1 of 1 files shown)


ankane:
Thanks @raulr!

PR #1508, opened 2021-08-09
Add support for different table/structure inherited models

rbviz:
Currently inherited models also inherit index name and reindex hooks (with searchkick re-define raising exception), but there's such use cases where inheritance is only code based and model hold different data structure, in which case searchkick simply does not support reindexing that model and fallbacks to it's superclass.

This PR tries to support different table indexes with inherited model code without breaking current inheritance options.

### DIFF
--- a/lib/searchkick/model.rb
+++ b/lib/searchkick/model.rb
@@ -22,24 +22,56 @@ module Searchkick
         raise ArgumentError, "Invalid value for callbacks"
       end
 
-      index_name =
-        if options[:index_name]
-          options[:index_name]
-        elsif options[:index_prefix].respond_to?(:call)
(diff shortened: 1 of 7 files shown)


jusmat (approved):
Looking good.

rbviz:
@ankane  I'm up for a different approach if needed, but supporting this code-only inheritance is essential for some of my projects. Right now trying to reindex a record of a model that inherits searchkick - reindexes whole parent scope and is a fatal bug in that configuration.

ankane:
Hey @rbviz, thanks for the PR. However, this isn't a feature I'd like to support right now.

rbviz:
This is regrettable, considering I'm offering to do all the implementation work.

Active record models that inherit other model code, but does not share the same database table - will remain broken with this stance.

Issue #1461, opened 2021-02-05
case_insenstive and flags options with regexp


Jaroost:
Is there a way to set [```case_insensitive``` and ```flags```](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-regexp-query.html) options when searching with a regexp?

ankane:
Hey @Jaroost, the Searchkick syntax doesn't support any flags right now, so you'd have to modify the query generated by Searchkick (see the Advanced Search section in the readme).

It looks like the `case_insensitive` option was recently added in Elasticsearch 7.10. I've added it to the plan for Searchkick 5, as the flag is ignored with a warning right now, so changing this would be a breaking change.

rordev-lab:
Hi @ankane 
Could you please let me know if the above case-insensitive-based searching is available now?
or if you could let me know that how could I get started to change Searchkick query syntex to get it worked? 
Thanks

jonowar:
@rordev-lab I had the same question and was playing around with it and reading the searchkick docs on [Advanced Search](https://github.com/ankane/searchkick#advanced-search) found this method modifying `body_options` to work:

```rb
Product.search(
  "francisco",
  body_options: {
    "query": {
      "bool": {
        "filter":[
          {
            "regexp":{
              "first_name":{
                "value":"J(I|A)M.I?E",
                "flags":"NONE",
                "case_insensitive": true
              }
            }
          }
        ]
      }
    }
  },
  load: false,
  limit: 1
)
```

Which will generate this elasticsearch query:

```
filed_contributions_production/_search
{
  "query":{
    "bool":{
      "should":[
        {
          "dis_max":{
            "queries":[
              {
                "multi_match":{
                  "query":"francisco",
                  "boost":10,
                  "operator":"and",
                  "analyzer":"searchkick_search",
                  "fields":["*.analyzed"],
                  "type":"best_fields"
                }
              },
              {
                "multi_match":{
                  "query":"francisco",
                  "boost":10,
                  "operator":"and",
                  "analyzer":"searchkick_search2",
                  "fields":["*.analyzed"],
                  "type":"best_fields"
                }
              },
              {
                "multi_match":{
                  "query":"francisco",
                  "boost":1,
                  "operator":"and",
                  "analyzer":"searchkick_search",
                  "fuzziness":1,
                  "prefix_length":0,
                  "max_expansions":3,
                  "fuzzy_transpositions":true,
                  "fields":["*.analyzed"],
[...]

PR #1510, opened 2021-08-12
Add support for case-insensitive regex filter

jonowar:
fixes https://github.com/ankane/searchkick/issues/1461

This PR will allow the use of case insensitive regexes in query filters.
If the `i` flag is present on the regex, then `case_insensitive` will be set to `true` in the filter, otherwise it will be set to `false`.

##### With case insensitive regex flag
```rb
3.0.2 :001 > Product.search('*', where: {name: /\Afoo\z/i})
  Product Search (25.3ms)  products_development/_search
{
  "query":{
    "bool":{
      "must":{
        "match_all":{}
      },
      "filter":[
        {
          "regexp":{
            "name":{
              "value":"foo",
              "flags":"NONE",
              "case_insensitive":true
            }
          }
        }
      ]
    }
  },
  "timeout":"16s",
  "_source":false,
  "size":10000
}
```

##### Without case insensitive regex flag
```rb
3.0.2 :002 > Product.kicksearch('*', where: {first_name: /\Afoo\z/})
  Product Search (22.2ms)  products_development/_search
{
  "query":{
    "bool":{
      "must":{
        "match_all":{}
      },
      "filter":[
        {
          "regexp":{
            "name":{
              "value":"foo",
              "flags":"NONE",
              "case_insensitive":false
            }
          }
        }
      ]
    }
  },
  "timeout":"16s",
  "_source":false,
  "size":10000
}
```

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -1028,10 +1028,6 @@ module Searchkick
       elsif value.nil?
         {bool: {must_not: {exists: {field: field}}}}
       elsif value.is_a?(Regexp)
-        if value.casefold?
-          Searchkick.warn("Case-insensitive flag does not work with Elasticsearch")
-        end
-
         source = value.source
         unless source.start_with?("\\A") && source.end_with?("\\z")
           # https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-regexp-query.html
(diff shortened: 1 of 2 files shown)


ankane:
Hey @jonowar, thanks for the PR! For Elasticsearch < 7.10, we should keep the warning (and add to the message "< 7.10"), and not pass `case_sensitive` since it'll cause errors.

ankane:
Looks great, thanks @jonowar!

jonowar:
Awesome! And thanks for the gem, @ankane !

PR #1518, opened 2021-09-25
README Update - Clarify how to work with multi_search

ramonpm:
Hey there, I recently had to refactor an application to use `multi_search` for better performance and noted the documentation wasn't very good at that point. I wish I had this info right away here, which would have saved me a couple of hours digging into this.

I initially created a Stackoverflow question about it https://stackoverflow.com/questions/69137572/how-to-use-searchkick-multi-search-along-with-with-highlights-undefined-met but after no response dug into the codebase and discovered the important parts.

So, here's my contribution to the documentation, let me know if anything seems off or what should be written differently.

Btw, Searchkick is awesome ;). Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1693,13 +1693,11 @@ coupons_query = Coupon.search("snacks", execute: false)
 Searchkick.multi_search([products_query, coupons_query])
 ```
 
-Then `products_query` and `coupons_query` will be an instance of `Searchkick::Query` where you can call:
-1. `products_query.results` -> for a Ruby `Array` of products
-2. `products_query.execute` -> for the regular `Searchkick::Result` result, don't worry about `#execute`, it won't make the Elasticsearch request again because the result is already stored in an instance variable of `products_query`.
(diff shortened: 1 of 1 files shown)


ankane:
Hey @ramonpm, thanks for the PR! It looks like some delegation methods like `with_highlights` were missing, and should be fixed in the commit above. Users shouldn't need to call `execute` directly.

ankane:
Also added a test case to catch this in the future.

ramonpm:
hey @ankane, what's the exact way to use multi_search + with_highlights then?

`search_query.with_highlights`? instead of `search_query.execute.with_highlights`? If that's the case I already have the PR updated.

ankane:
Yeah, that's the recommended way. The readme currently has "Then use `products` and `coupons` as typical results.", which should now work with all methods, so I don't think an update is needed.

ramonpm:
ok, your call, closing the PR then. Thanks.

PR #1521, opened 2021-10-12
README Update - Clarify how to search by multi_locations

ZeeshanAslam005:
Hey there,
I was recently writing a logic to filter by multiple locations. I did not seem in documentation to search by multiple locations. I wish I had this info right away in here, which would have saved me a couple of hours digging into this.

So, here's my contribution to the documentation, let me know if anything seems off or what should be written differently.

Thank you.

Thanks for contributing. You’re awesome! A few things to keep in mind:

- Keep changes to a minimum
- Follow the existing style
- Add one or more tests if possible

Finally, replace all this with a description of the changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -943,6 +943,12 @@ Bounded by a polygon
 Restaurant.search "dessert", where: {location: {geo_polygon: {points: [{lat: 38, lon: -123}, {lat: 39, lon: -123}, {lat: 37, lon: 122}]}}}
 ```
 
+Search By Multiple Locations:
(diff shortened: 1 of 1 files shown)


ankane:
Hey @ZeeshanAslam005, thanks for the PR, and sorry for the delay. `_or`, `_and`, and `_not` can be used with any `where` options, and I'm not sure this is very common, so it's not something I'd like to include in the readme (but this should help anyone searching for it).

PR #1525, opened 2021-11-04
Adds entry in README for mixing matcher.

LeFnord:


LeFnord:
will close in favour of a post → [How To … Searchkick](http://www.le-fnord.eu/How_To_-_Searchkick/)

PR #1527, opened 2021-11-08
Add support for beginless ranges in where queries

### LINKED ISSUE #1524: Support for beginless ranges in where queries
I was trying to perform a query where a specific field was less than a given value. I knew that endless ranges have been supported in searchkick for a while, so I tried to use a beginless range:

```rb
Model.search(where: { field: (..10) })
```

But this raises an error in the Query class because it tries to call `first` on the beginless range which causes a RangeError.

I patched this feature in to my own fork and I would be happy to open a PR for it if you're interested in supporting this use case. 

I wrote a few different versions, I don't know which one most matches the style and tradeoffs you prefer for this project:

```rb
# A
def expand_range(range)
  expanded = {}
  expanded[:gte] = range.begin if range.begin
  if range.end && !(range.end.respond_to?(:infinite?) && range.end.infinite?)
    expanded[range.exclude_end? ? :lt : :lte] = range.end
  end
  expanded
end

# B
def expand_range(range)
  {
    gte: range.begin,
    (range.exclude_end? ? :lt : :lte) =>
      (range.end unless range.end.respond_to?(:infinite?) && range.end.infinite?),
  }.compact
end

# C
def expand_range(range)
  {
    gte: range.begin,
    (range.exclude_end? ? :lt : :lte) =>
      (range.end unless range.end.respond_to?(:infinite?) && range.end.infinite?),
  }.tap(&:compact!)
end
```

I haven't run benchmarks, but I believe A would be the fastest, but isn't the nicest looking to read to my eye. B is probably my personal preference, but it has an extra Hash allocation to run the `compact`. And C is an optimization of B with the sacrifice of some clarity by the added complexity of the `tap`.

If you're open to this suggestion, I can open a PR and add some tests for this.

willcosgrove:
This PR adds support for beginless ranges (`..5`) in the `where` field of the search method.

Closes #1524

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -904,12 +904,7 @@ module Searchkick
         else
           # expand ranges
           if value.is_a?(Range)
-            # infinite? added in Ruby 2.4
-            if value.end.nil? || (value.end.respond_to?(:infinite?) && value.end.infinite?)
-              value = {gte: value.first}
-            else
-              value = {gte: value.first, (value.exclude_end? ? :lt : :lte) => value.last}
-            end
+            value = expand_range(value)
           end
 
           value = {in: value} if value.is_a?(Array)
@@ -1136,6 +1131,17 @@ module Searchkick
       else
         value
       end
+    end
+
+    def expand_range(range)
+      expanded = {}
+      expanded[:gte] = range.begin if range.begin
+
+      if range.end && !(range.end.respond_to?(:infinite?) && range.end.infinite?)
+        expanded[range.exclude_end? ? :lt : :lte] = range.end
+      end
+
+      expanded
     end
 
     def base_field(k)
--- a/test/where_test.rb
+++ b/test/where_test.rb
@@ -46,6 +46,11 @@ class WhereTest < Minitest::Test
       # use eval to prevent parse error
       assert_search "product", ["Product C", "Product D"], where: {store_id: eval("3..")}
     end
+    if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.7.0")
+      # use eval to prevent parse error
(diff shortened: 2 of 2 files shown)


lib/searchkick/query.rb
```diff
@@ -1138,6 +1133,17 @@ def location_value(value)
...
 
+    def expand_range(range)
+      expanded = {}
+      expanded[:gte] = range.begin if range.begin
+
+      if range.end && !(range.end.respond_to?(:infinite?) && range.end.infinite?)
+        expanded[range.exclude_end? ? :lt : :lte] = range.end
+      end
+
+      expanded
```
willcosgrove, line 1144: I considered using `Object#tap` here but after benchmarking it, decided it wasn't worth the (albeit small) performance hit.

```rb
require "benchmark/ips"

def explicit_return
  hash = {}

  hash[:some_value] = "some_string"

  hash
end

def tap_implicit_return
  {}.tap do |hash|
    hash[:some_value] = "some_string"
  end
end

Benchmark.ips do |b|
  b.report("explicit return") { explicit_return }
  b.report("tap implicit return") { tap_implicit_return }

  b.compare!
end
```

```
Warming up --------------------------------------
     explicit return     1.042M i/100ms
 tap implicit return   816.126k i/100ms
Calculating -------------------------------------
     explicit return     10.519M (± 0.9%) i/s -     53.152M in   5.053522s
 tap implicit return      8.177M (± 0.9%) i/s -     41.622M in   5.090811s

Comparison:
     explicit return: 10518686.7 i/s
 tap implicit return:  8176709.5 i/s - 1.29x  (± 0.00) slower
```

ankane:
Looks great, thanks @willcosgrove! (fwiw, I personally find the submitted/non-tap version cleaner)

PR #1536, opened 2022-01-18
docs: "complete list of analyzers." use relative file link

nbw:
The link to the analyzers is a hardcoded link to a previous version of the file. Instead, use a relative link that points to the most recent version. 

I noticed this because all of the language analyzers are in not in the old link.


ankane:
Hey @nbw, thanks for reporting and the PR! Unfortunately, the line number will be off with a relative link, but I've updated it to the latest version.

ankane:
Sorry for the noise. On second thought, probably better that than the list of analyzers being outdated.

PR #1562, opened 2022-05-21
Adds bulk_queue_name as an option

weilandia:
This PR adds a `bulk_queue_name` option for `BulkReindexJob`. This would be helpful as currently when we bulk reindex the `searchkick` queue gets filled up and delays incoming (time sensitive) reindex jobs.

There may already be a better solution for this as I imagine others face the same issue, but I didn't see anything.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 5.0.4 (unreleased)
 
 - Added `max_result_window` option
+- Added `bulk_queue_name` option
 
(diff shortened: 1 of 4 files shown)


ankane:
Hey @weilandia, thanks for the PR. You can create an initializer with:

```ruby
Searchkick::BulkReindexJob.queue_as :bulk_search_reindex
```

so I don't think an additional option is needed.

Issue #1549, opened 2022-03-07
Support range as an option for async reindex


kapso:
I am curious why this change was made? - `Changed async reindex to fetch ids instead of using ranges for numeric primary keys with Active Record`

...and if it's possible (or perhaps makes sense) to support range as an option as well for async reindex

Why? - enqueuing ids is really slow, especially when writing 100s of ids/job to Sidekiq/Redis. We use a `batch_size` of 1000 and creating jobs is really slow when reindexing 40m documents.

**UPDATE:** it took ~16 hours to enqueue ids for 40m documents with a batch size of 1000.

kapso:
Or maybe use `Sidekiq::Worker.perform_bulk` which I think could speed up Sidekiq/Redis writes, in case Sidekiq is being used.

kapso:
@ankane any thoughts? thanks.

ankane:
Hey @kapso, I was thinking it'd be good to handle gaps in primary keys at the expense of longer enqueuing times, but may be good to support the previous pattern as well. How long did it previously take to enqueue, and how does the total reindexing time compare before and after?

kapso:
@ankane We have `140m` rows in the table, out of which 40m rows (we use `search_import` scope for filtering) get indexed. With a `batch_size` of 500, it takes 20+ hours to just enqueue all ids.

The previous pattern used to take ~20m.

We use the Heroku worker to enqueue these ids, and sometimes Heroku workers get re-cycled and so does the enqueue process.

kapso:
@ankane another interesting thing we have seen a few times is that the Enqueuing job (Heroku worker) crashes after a couple of hours with no error reported in our error system (Sentry). And since it crashed (and re-started) it obviously re-started the enqueuing process - creating a fresh index and writing to this new index.

The  Enqueuing (Heroku) worker had 2.5GB memory, so memory definitely wasn't an issue.

This is the enqueuing job/process - `Model.reindex(mode: :async)`

kapso:
@ankane curious if there are any plans to also support the range option?

kapso:
@ankane thinking of submitting a PR for this, is this the place to start? - https://github.com/ankane/searchkick/commit/88f52da847c1a8c7d034ffe3caec66098cf431db

Or is there any other code we should also be looking at?

BobbyMcWho:
```ruby
if relation.respond_to?(:find_in_batches)
  relation.find_in_batches(batch_size: batch_size) do |items|
[...]

PR #1571, opened 2022-06-29
[WIP] Use pluck instead of map

BobbyMcWho:
Use in_batches.each with pluck in the AR case, so that we avoid loading all of the objects into memory when we're really just looking for ids here. 

Mongoid supports pluck as of 3.0.1. I'm not sure what version searchkick targets, but it looks like latest Mongoid is 5.0+, so maybe worth dropping support for lower than 3.0.1? 

This is a PR after a cursory read of the code and reading #1549, so please challenge any assumptions I've made. 

I couldn't get tests running locally, it was unclear in the README#contributing section, so I may just not have ES running locally. 
```ruby
The client is unable to verify that the server is Elasticsearch. Some functionality may not be compatible if the server is running an unsupported product.
/Users/bobbymcdonald/.asdf/installs/ruby/3.0.3/lib/ruby/gems/3.0.0/gems/typhoeus-1.4.0/lib/typhoeus/adapters/faraday.rb:106:in `block in request': Server returned nothing (no headers, no data) (Faraday::ConnectionFailed)
```

### DIFF
--- a/lib/searchkick/relation_indexer.rb
+++ b/lib/searchkick/relation_indexer.rb
@@ -63,10 +63,10 @@ module Searchkick
     end
 
     def in_batches(relation)
-      if relation.respond_to?(:find_in_batches)
+      if relation.respond_to?(:in_batches)
         klass = relation.klass
         # remove order to prevent possible warnings
-        relation.except(:order).find_in_batches(batch_size: batch_size) do |batch|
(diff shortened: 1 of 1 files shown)


ankane:
Hey @BobbyMcWho, thanks for the PR. However, the full object is needed in some cases, and this isn't something I'd like to spend time on right now.

PR #1582, opened 2022-09-12
Add model name to warning message

sasharevzin:
I find `[searchkick] WARNING: Records in search index do not exist in database: 4598, 5450, 6192, 6216, 7018, 7245, 7266` unclear to which model these ids related. For better understanding, I would like to add the model name so that it will look like `[searchkick] WARNING: User records in search index do not exist in the database: 4598, 5450, 6192, 6216, 7018, 7245, 7266`

### DIFF
--- a/lib/searchkick/results.rb
+++ b/lib/searchkick/results.rb
@@ -302,7 +302,7 @@ module Searchkick
     def build_hits
       @build_hits ||= begin
         if missing_records.any?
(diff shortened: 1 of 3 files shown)


ankane:
Hey @sasharevzin, thanks for the suggestion and PR. Added in the commit above (there can be multiple model names in a result).

PR #1584, opened 2022-09-19
Allow to skip reindex

kshnurov:
`searchkick` triggers reindex on any change, including touch, but some fields (e.g. `updated_at`) may not be stored in index and there's no point in reindexing.

Current docs suggest manually defining callbacks, but that won't work with `Searchkick.callbacks(:bulk) do`, which will always trigger a reindex.

I've added a `skip_searchkick_reindex?` method which users may override to define their own logic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2058,10 +2058,11 @@ Reindex conditionally
 
 ```ruby
 class Product < ApplicationRecord
-  searchkick callbacks: false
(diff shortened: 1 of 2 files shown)


kshnurov:
@ankane could you please review this?

kshnurov:
@ankane ?

PR #1589, opened 2022-09-26
Resolve redis 5.0 deprecation warning

### LINKED ISSUE #1588: Redis deprecation warning
Running a batch reindex raises a deprecation warning 
```
Redis::Deprecated: Redis#sadd will always return an Integer in Redis 5.0.0. Use Redis#sadd? instead.(called from: /usr/local/bundle/gems/searchkick-5.0.4/lib/searchkick/relation_indexer.rb:137:in `block in batch_job')
```

BobbyMcWho:
resolves #1588

### DIFF
--- a/lib/searchkick/relation_indexer.rb
+++ b/lib/searchkick/relation_indexer.rb
@@ -134,7 +134,7 @@ module Searchkick
     end
 
     def batch_job(class_name, batch_id, record_ids)
-      Searchkick.with_redis { |r| r.sadd(batches_key, batch_id) }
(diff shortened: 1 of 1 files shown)


ankane:
Thanks @BobbyMcWho! Just pushed a similar change for `srem` as well.

A few notes to self for future reference:
- the warning only appears in redis 4.8 gem
- redis 3 gem (released 2012) and beyond supports passing an array to `sadd`/`srem`
- redis 2 gem accepts an array, but stores it as a string (`[value]`, which is fine since it applies to both commands)

PR #1594, opened 2022-10-24
Allow string options' keys

oz-tal:
I have use case where my options come from a JSON object which might, or might not, be for Searchkick's search method. In this scenario, converting the keys/values to symbols/integers is impractical. I'd like to be able to pass them as is, and let Searchkick cast them to the proper format. 

As far as I can see, there's very little downside in enabling such behavior. This PR enable calling options with string keys and allow passing the offset as a string, eg. : 
```ruby
User.search('*', 'limit' => '12', 'offset' =>  '24')
```
Which otherwise would required to be:
```ruby
User.search('*', :limit => 12, :offset =>  24)
```

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -17,6 +17,8 @@ module Searchkick
       :with_score, :misspellings?, :scroll_id, :clear_scroll, :missing_records, :with_hit
 
     def initialize(klass, term = "*", **options)
+      options.transform_keys!(&:to_sym)
+
       unknown_keywords = options.keys - [:aggs, :block, :body, :body_options, :boost,
         :boost_by, :boost_by_distance, :boost_by_recency, :boost_where, :conversions, :conversions_term, :debug, :emoji, :exclude, :explain,
         :fields, :highlight, :includes, :index_name, :indices_boost, :limit, :load,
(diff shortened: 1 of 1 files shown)


ankane:
Hey @oz-tal, thanks for the PR. Since users can call `symbolize_keys` on options before passing them to Searchkick, I don't think additional functionality is needed.

oz-tal:
Hi @ankane, thanks for taking the time to reply, I'd like to argue my case a little bit further.

While you're right in saying the user can call `symbolize_keys` himself, consider this:
```ruby
# valid calls
User.search('*', :limit => 12, :offset =>  24)
User.search('*', :limit => '12', :offset =>  24)
# raise when the offset is used
User.search('*', :limit => '12', :offset =>  '24')
# raise as 'invalid option'
User.search('*', 'limit' => '12', 'offset' =>  '24')
```
It's unintuitive for the user to have the bottom two calls blow up, while the top two work. It's also not necessary immediately clear why from the returned errors. I'd also like to state the following:

* Casting string keys to symbols cannot cause unexpected side effects.
* The user's intention behind passing a string key and a symbol key will always be the same.
* The added complexity to the code is minimal.
* It's a small layer of flexibility.
* The performance hit is highly negligible.
* Other major widespread gems like ActiveRecord and Mongoid support the dynamic casting of strings in similar context. This reinforce the behavior as expected by the user.

If you're still not convinced, can you consider at least casting the offset to an integer? It would be consistent with the behavior for the `limit`/`per_page` options here: https://github.com/ankane/searchkick/blob/master/lib/searchkick/query.rb#L252

ankane:
Hey @oz-tal, I think casting `offset` makes sense (feel free to create a new PR), but don't think it's common for methods to accept keyword arguments as both symbols and strings.

oz-tal:
@ankane Alright, thanks you for your feedback! I opened another PR including only the offset casting.

PR #1596, opened 2022-11-07
Cast offset option to integer

oz-tal:
Cast the offset option to allow passing the value as a string in the following fashion:
```ruby
model.search('*', limit: '12', offset: '24')
```


ankane:
Thanks @oz-tal!

PR #1598, opened 2022-11-10
Fixed typo in circle-based geo-shape query in readme

buffpojken:
There's a typo in the query-format for circle-based geo-shape queries, which expects a single point, not an array.


ankane:
Thanks @buffpojken!

buffpojken:
My pleasure!

PR #1600, opened 2022-12-06
Add support for script filtering with painless

AlecR:
This PR enables support for the `_script` filter in Searchkick queries. I would like to add this functionality to support more complex querying cases where the base query language is insufficient.

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -904,8 +904,8 @@ module Searchkick
           filters << {bool: {must_not: where_filters(value)}}
         elsif field == :_and
           filters << {bool: {must: value.map { |or_statement| {bool: {filter: where_filters(or_statement)}} }}}
-        # elsif field == :_script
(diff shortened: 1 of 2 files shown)


sjl2 (commented):
Is this risky in terms of security or performance?

AlecR:
In terms of security, Elastic has built in security layers to ensure scripts are executed safely. [These Elastic docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting-security.html) give a good overview of the layers of security they've implemented.

Performance of scripts is going to be mostly case-dependent, as the size of the data queried and amount of computation required can vary widely. Adding support for this feature will not inherently impact the performance of searchkick queries, but writing an inefficient script certainly would.

nick-galluzzo (approved):

ankane:
Hi @AlecR, thanks for the PR. Added support for this in the commit above.

Issue #1141, opened 2018-05-10
Use set over list on ReindexQueue#push


dvmonroe:
First, thanks for all your work @ankane ❤️ 

Curious what you're thoughts were behind storing record ids in a list vs a set at: https://github.com/ankane/searchkick/blob/17c9adae455fb600f847351b0f4ab15d87ad1092/lib/searchkick/reindex_queue.rb#L12

Time complexity? Something else?

It threw me a little when I used the length method like so:

```ruby
User.search_index.reindex_queue.length
```
And it was greater than the number of actual User records that had been updated in our db and eventually reindexed (due to a few records being updated twice within an x amount of time).

Is this worth a PR?

dvmonroe:
Thanks @ankane. I gotcha. Perhaps I'm pulling at straws but the api seems less than ideal given a scenario like so:

```ruby
# app/models/post.rb
# == Schema Information
#
# Table name: posts
#
#  id 
#  likes

Post < ApplicationRecord
  # unrealistic batch size for demo purposes
  searchkick callbacks: :queue, batch_size: 5
end
```

Suppose a Post record gets updated everytime a user hits "like" via the UI (replicated in pry):
```pry
> (0..50).each { |i| Post.first.update(likes: i) };
> Post.search_index.reindex_queue.length
50
> $redis.lrange("searchkick:reindex_queue:posts", 0, 4)
["1","1","1","1","1"] # these are the ids that will rpop on Searchkick::ReindexQueue#reserve
```

```pry
lib/searchkick/process_batch_job.rb#perform
> Searchkick.load_records(Post, ["1","1","1","1","1"]).length
1
```
**Main isssues**
1)  I'd want a cron job to handle the reindex once I have x number of unique post records in the queue but the current searchkick api wouldn't tell me that

2) In this scenario, `Searchkick::ProcessQueueJob.perform_later(class_name: "Post")` would reindex my first Post record 10 times. But the searialized JSON of the Post record wouldn't be any different in each of those elastic api _bulk calls.

**Potential solution**

Only queueing the post once would appear to be more performant for the ES writes (via a set (sadd) and/or ordered set (zadd))

Perhaps, I'm missing the benefit of being able to queue the same record multiple times and/or I'm making mountains out of molehills i.e this would see very little/no performance benefits on the ES cluster in which case - feel free to close!

Thanks again.

ankane:
[...]

PR #1602, opened 2022-12-07
Replace Redis lists usage for Redis sets for reindex queue

dewski:
This PR updates Searchkick::ReindexQueue to use [Redis sets](https://redis.io/docs/data-types/sets/) instead of lists when keeping track of records that need to be reindexed.

> A Redis set is an unordered collection of unique strings (members). You can use Redis sets to efficiently:
> - **Track unique items (e.g., track all unique IP addresses accessing a given blog post).**
> - Represent relations (e.g., the set of all users with a given role).
> - Perform common set operations such as intersection, unions, and differences.

This change will prevent the reindex queue from ballooning with duplicate records at risk of increasing Redis memory usage and unnecessary duplicate reindexing as it works through the list.

## Background

I observed an application on Heroku using the [Heroku Redis Premium 0](https://elements.heroku.com/addons/heroku-redis) plan running into out of memory errors all of a sudden for the 50 MB allocated to the Redis instance. Upon further investigation using `redis-cli --bigkeys` I found 2 of my reindex queues were taking up 48 MB of the 50 MB.

Inspecting an example range of records needing to be reindexed, only 21 out of the 1,000 records were unique.

```ruby
ids = Stock.redis.lrange("searchkick:reindex_queue:products_production", 0, 999)
ids.uniq.length # => 21
ids.length # => 1000
```

Redis information:

```
> LLEN "searchkick:reindex_queue:products_production"
(integer) 6176148
> INFO
# Memory
used_memory:52028680
used_memory_human:49.62M
> DEL "searchkick:reindex_queue:products_production"
(integer) 1
> INFO
# Memory
used_memory:27195120
used_memory_human:25.94M
> DEL "searchkick:reindex_queue:product_variants_production"
(integer) 1
> INFO
# Memory
used_memory:2424296
used_memory_human:2.31M
```

After the Redis instance ran out of memory, no more jobs could be enqueued. Most notably the `Searchkick::ProcessQueueJob` which would have worked off some of the records queued to be reindexed. For my particular setup, I have the `Searchkick::ProcessQueueJob` job going every 10 minutes (due to Heroku Scheduler having 10 minutes be the lowest interval offered).

## Breaking change

One risk with this change is it would be considered a breaking change as the Redis key would have to be updated to use sets in place of lists to account for existing reindex queues upon a version upgrade.

```
> SPOP "searchkick:reindex_queue:products_production" 999
(error) WRONGTYPE Operation against a key holding the wrong kind of value
```

**Changing the reindex queue Redis key name**
This would avoid any Redis errors, but risks records queued to be reindexed never actually reindexed. Instructing users how to migrate in a major release bump or provide a Rake task could suffice (untested illustrative script):

```ruby
def redis_version
  @redis_version ||= Searchkick.with_redis { |r| Gem::Version.new(r.info["redis_version"]) }
end

def supports_rpop_with_count?
  redis_version >= Gem::Version.new("6.2")
end

def reserve(redis_key, limit:)
  record_ids = if supports_rpop_with_count?
    Searchkick.with_redis { |r| r.call("rpop", redis_key, limit) }.to_a
  else
    record_ids = []
    Searchkick.with_redis do |r|
      while record_ids.size < limit && (record_id = r.rpop(redis_key))
        record_ids << record_id
      end
    end
    record_ids
  end
end

def migrate(model, limit: 1000)
  redis_key = "searchkick:reindex_queue:#{model.search_index.name}"
  
  while (record_ids = reserve(redis_key, limit: limit)).any?
    model.reindex_queue.push(record_ids)
  end

  Searchkick.with_redis do |r|
    r.del(redis_key)
  end
end

Searchkick.models.each do |model|
  migrate(model)
end
```

### DIFF
--- a/lib/searchkick/reindex_queue.rb
+++ b/lib/searchkick/reindex_queue.rb
@@ -50,14 +50,6 @@ module Searchkick
       "searchkick:reindex_queue:#{name}"
     end
 
-    def supports_rpop_with_count?
-      redis_version >= Gem::Version.new("6.2")
-    end
-
-    def redis_version
-      @redis_version ||= Searchkick.with_redis { |r| Gem::Version.new(r.info["redis_version"]) }
-    end
-
     def escape(value)
       value.gsub("|", "||")
     end
--- a/lib/searchkick/reindex_queue.rb
+++ b/lib/searchkick/reindex_queue.rb
@@ -10,7 +10,7 @@ module Searchkick
 
     # supports single and multiple ids
     def push(record_ids)
-      Searchkick.with_redis { |r| r.lpush(redis_key, record_ids) }
+      Searchkick.with_redis { |r| r.sadd(redis_key, record_ids) }
     end
 
     def push_records(records)
@@ -33,17 +33,7 @@ module Searchkick
 
     # TODO use reliable queuing
     def reserve(limit: 1000)
-      if supports_rpop_with_count?
-        Searchkick.with_redis { |r| r.call("rpop", redis_key, limit) }.to_a
-      else
-        record_ids = []
-        Searchkick.with_redis do |r|
-          while record_ids.size < limit && (record_id = r.rpop(redis_key))
-            record_ids << record_id
-          end
-        end
-        record_ids
-      end
+      Searchkick.with_redis { |r| r.spop(redis_key, limit) }
     end
 
     def clear
@@ -51,7 +41,7 @@ module Searchkick
     end
 
     def length
-      Searchkick.with_redis { |r| r.llen(redis_key) }
+      Searchkick.with_redis { |r| r.scard(redis_key) }
     end
 
     private
--- a/test/reindex_test.rb
+++ b/test/reindex_test.rb
@@ -47,6 +47,22 @@ class ReindexTest < Minitest::Test
     assert_search "product", ["Product A"]
   end
 
+  def test_record_queue_does_not_grow_on_duplicate_records
+    reindex_queue = Product.search_index.reindex_queue
+    reindex_queue.clear
+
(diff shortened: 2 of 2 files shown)


ankane:
Hey @dewski, thanks for the PR, but check out #1141 and #1495.

dewski:
> Hey @dewski, thanks for the PR, but check out #1141 and #1495.

Would you be open to an alternative PR that supports 2 queue strategies, one that is ordered and uses Redis lists (default) and another that is unique and uses Redis sets (configurable)?

```ruby
  searchkick \
    callbacks: :queue,
    batch_size: 10,
    queue: :unique
end
```

Following the best practices listed in the README to batch index updates, order doesn't particularly matter for us. The default behavior grew the queue too large and resulted in Redis running out of memory. In our particular case, our search document contains data about many relationships and needs to be updated if any of the references records are updated (name, description, inventory, pricing).

In the same PR I would also update the documentation to reference the pros and cons of each strategy. For our situation, the pros/cons are below:

### Redis Lists

**Pros**

- Ordered processing

**Cons**

- Can contain duplicate records resulting in extra batches being performed
- A given record can be indexed N times in a reindexing job depending on the configured batch size and how many times the duplicate record ID appears in the batch
  - In our case we accumulated 6M+ records being added to the reindex queue list while we only have 5K records in total with many `has_many` relationships touching the referenced record causing the duplication. We have configured the queue reindex batch size to 10 due to the large ES document size it generates.

### Redis Sets

**Pros**

- Unique set of records being updated
- Same record is indexed only once during a reindex job

**Cons**

- Unordered processing

ankane:
It's not something I'd like to maintain, but someone could create a separate gem for this pattern.

Issue #1288, opened 2019-07-15
Searchkick 5


ankane:
Please create a new issue to discuss any ideas.

`edge` branch

New

- [x] Add support for Elasticseach 8 (works w/o any changes)
- [x] Support `mode: :async` and `mode: :queue` with bulk reindexing

Breaking

- [x] Remove `elasticsearch` dependency from gemspec
- [x] Drop support for Elasticsearch 6
- [x] Drop support for NoBrainer and Cequel
- [x] Drop support for deprecated aws_signers_v4 (prefer aws_sigv4 instead)
- [x] Lazy load queries
- [x] Raise `ArgumentError` instead of `RuntimeError` for unknown operators
- [x] Add `.*` to non-anchored regular expressions
- [x] Raise error when `search` called on relations
- [x] Fix or remove `wordnet` option (could remove first and re-add later)
- [x] Use `like: [{_index: ..., _id: ...}]` for similar records
- [x] Replace `japanese` with `japanese2`
- [x] No longer map `id` to `_id` with `order` option (since sorting on `_id` is deprecated in Elasticsearch)
- [x] Raise an error for unpermitted parameters (like Active Record)
- [x] Raise `ArgumentError` (instead of warning) for invalid regular expression modifiers
- [x] Have relation reindex remove records when `should_index?` is `false` - #1424

Other

- [x] Use module to prevent `method redefined; discarding old search_data` warnings
- [x] Use `ActiveSupport.on_load(:mongoid)` for Mongoid
- [x] Drop support for Ruby < 2.6 and Active Record < 5.2
- [x] Lazy load index name - `lazy_index_name` branch
- [x] Fix `scope_attributes` for relation reindex - `scope_attributes` branch
- [x] Decide how to handle fields that use Searchkick options like `word_start` and custom settings with `merge_mappings` (throw error?) - #1410
- [x] Decide how to handle elasticsearch-ruby 8 response object - https://www.elastic.co/guide/en/elasticsearch/client/ruby-api/8.0/release_notes_80.html

Maybe

- [ ] Better notifications - log data on store and update and add filtering for sensitive data (similar to Active Record [filter_attributes](https://blog.bigbinary.com/2019/09/03/rails-6-adds-activerecord-base-filter_attributes.html))
- [ ] Have `relation.reindex` load all records - use `relation.in_batches.reindex` to load in batches
- [ ] `reindex_now`/`reindex_later` methods
- [x] Add `with_score` method to `Searchkick::Results`
[...]

PR #1617, opened 2023-03-09
Restore WordNet support

sharshenov:
Apparrently, WordNet synonyms just work when `"format": "wordnet"` param is given.

References:
- https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-synonym-graph-tokenfilter.html#_wordnet_synonyms_2
- https://github.com/ankane/searchkick/issues/1288 (WordNet support was dropped for Searchkick 5.0)

### DIFF
--- a/README.md
+++ b/README.md
@@ -410,6 +410,27 @@ And reload with:
 Product.search_index.reload_synonyms
 ```
 
+#### WordNet
+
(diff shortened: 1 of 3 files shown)


ankane:
Hi @sharshenov, thanks for the PR. However, I don't think this feature is very common (it was broken for a while and no one reported), so it's not something I'd like to restore.

PR #1618, opened 2023-03-24
Added arguments to unscoped option

marelons1337:
This is a proposition for my specific use case, but I realized that could be useful regardless.
My current use case:
```ruby
class Person
  searchkick word_start: [:name]

  default_scope -> { where(account_id: current_account) }
  default_scope -> { where.not(kind: 'guest') }
  
  scope :guests, -> { rewhere(kind: 'guest') }
end
```

Now, normally I don't need to see guests, but when searching with searchkick, I'd like to see all records.
However if I choose to use 
```ruby
  searchkick word_start: [:name], unscope: true
```
it would also unscope `account_id` and this is a scope I need every time.
Which brings us to this proposition, rather simple, but I feel like it was a missing feature. It's my first time contributing so please let me know if you have any questions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -664,6 +664,16 @@ class Product < ApplicationRecord
 end
 ```
 
+If you would like to only exclude certain default scopes, set:
+```ruby
+class Person
+  searchkick word_start: [:name], unscope: { where: :kind }
+
+  default_scope -> { where(account_id: current_account) }
+  default_scope -> { where.not(kind: 'guest') }
+end
+```
+
 ## Intelligent Search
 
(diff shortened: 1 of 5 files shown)


zrizvi1 (approved):

ankane:
Hi @marelons1337, thanks for the PR, including tests! You should be able to re-add scopes through `search_import`, so I don't think additional complexity is needed.

marelons1337:
Hi @ankane, thanks for looking into it! I just realised I made a silly mistake since I'm using a scope with `current_account` when it's available, but of course it's not when I'm reindexing the data. I'm still going with `unscope` but sticking to filtering the data after I query it from elastic.
Might've been a waste of time, but it's good to know these PR's are looked into, thanks a lot! 👍

Issue #1481, opened 2021-04-07
Support for Count API 


kapso:
Add support for Count API - https://www.elastic.co/guide/en/elasticsearch/reference/current/search-count.html

I had asked this question here - https://stackoverflow.com/questions/63002404/searchkick-count-query

ankane:
Hey @kapso, thanks for the suggestion. I've considered this in the past (#632), but don't think it's common enough to support now. You may be able to get similar results with `Model.search("query", limit: 0).total_count`.

kapso:
@ankane thanks, that makes sense. Guessing also need to add `body_options: { track_total_hits: true }`

PR #1619, opened 2023-03-27
Update Index#total_docs to pull in full record count

JeremiahChurch:
Referencing https://github.com/ankane/searchkick#deep-paging - update the total_docs method to pull in the full count of records instead of the 10k default limited count.

This was mentioned in #1481 

endless ❤️  for searchkick


ankane:
Ah, looks like that was broken for a while. Thanks @JeremiahChurch!

PR #1621, opened 2023-04-03
Better dependency injection on Index and Indexer

ombr:
While trying to use multiple Elasticsearch servers, we realize Searchkick uses a single Indexer that uses the default client.
We just changed the indexer to have the client injected and the index to use a single function to get the indexer.


ankane:
Hi @ombr, thanks for the PR, but don’t plan to support any changes to internal APIs.

PR #1627, opened 2023-04-17
cast value in escape to a string

### LINKED ISSUE #1626: ReindexQueue#push_records doesn't like routing to be non-string value
**Describe the bug**

https://github.com/ankane/searchkick/blob/master/lib/searchkick/reindex_queue.rb#L27 expects `routing`/`record.search_routing` to be a string. in the case of a numeric the gsub call in the escape method raises a NoMethodError.
**To reproduce**
contrived example using the standard example:

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"

  gem "activerecord", require: "active_record"
  gem "activejob", require: "active_job"
  gem "sqlite3"
  gem "searchkick", git: "https://github.com/ankane/searchkick.git"
  gem "elasticsearch", '< 7.14'
  gem 'redis'
end

puts "Searchkick version: #{Searchkick::VERSION}"
puts "Server version: #{Searchkick.server_version}"

ActiveRecord::Base.establish_connection adapter: "sqlite3", database: ":memory:"
ActiveJob::Base.queue_adapter = :inline

ActiveRecord::Schema.define do
  create_table :products do |t|
    t.string :name
  end
end

class Product < ActiveRecord::Base
  searchkick callbacks: :queue, routing: true

  def search_routing
    id
  end
end

Searchkick.redis = Redis.new
Product.reindex
Product.create!(name: "Test")
=> /lib/searchkick/reindex_queue.rb:77:in `escape': undefined method `gsub' for 1:Integer (NoMethodError)
```

**Additional context**
Numeric IDs otherwise seem to work good in search routing. I'm not sure what the right/preferred solution is. Should the `search_routing` method always be cast to a string? or should push_records pass `routing.to_s` to the escape method? or should escape to_s the value passed in? If you can point me which direction you'd prefer for the patch I'm happy to pull a PR together for it.

We're running with the 3rd option right now in prod with no issues but we're only hours out from moving from async to queues
```
def escape(value)
  value.to_s.gsub("|", "||")
end
```

JeremiahChurch:
fixes #1626

### DIFF
--- a/lib/searchkick/reindex_queue.rb
+++ b/lib/searchkick/reindex_queue.rb
@@ -74,7 +74,7 @@ module Searchkick
     end
 
     def escape(value)
-      value.gsub("|", "||")
+      value.to_s.gsub("|", "||")
     end
   end
 end


ankane:
Thanks @JeremiahChurch!

PR #1643, opened 2023-08-01
Do not reindex type automatically.

vgsantoniazzi:
When doing a partial reindex (to index a new field, for example), searchkick was reindexing `type`  even when not asked to.

This is a problem because maybe you want to reindex `type` with a different value, and after doing a partial index, you need to call the `reindex` again by type — mainly because searchkick has overridden the type using the `document_type` method.

Performance is a concern here. Imagine doing that for millions of rows; it will take more time to finish the reindexing.


vgsantoniazzi:
@ankane can we get a review here?

vgsantoniazzi:
@ankane :D

PR #1645, opened 2023-08-30
Change the default value of fuzziness to `AUTO`

kzkn:
`AUTO` is the preferred value of `fuzziness`.

ref: https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#fuzziness

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -317,7 +317,7 @@ module Searchkick
           end
 
           if misspellings != false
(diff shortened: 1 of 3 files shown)


ankane:
Hi @kzkn, thanks for the PR. However, this is a breaking change, and I think the current default is fine.

kzkn:
@ankane 
Thanks for the comment :smiley: 

What do you think about making configurable the default value of `fuzziness`? For the compatibility, the default value of the configuration will be 1.

ankane:
I don't think an additional option is needed.

PR #1655, opened 2023-11-04
Don't search misspellings for an errored query

davidrunger-cl:
Hello, and thank you for this great library!

When executing a multisearch query that ultimately causes Elasticsearch to error and when using the misspellings option, we are seeing an unhandled exception bubble up from `searchkick`:

```
NoMethodError:
  undefined method `[]' for nil:NilClass
# /Users/david/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/searchkick-5.3.0/lib/searchkick/results.rb:92:in `total_count'
# /Users/david/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/searchkick-5.3.0/lib/searchkick/query.rb:194:in `retry_misspellings?'
# /Users/david/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/searchkick-5.3.0/lib/searchkick/multi_search.rb:22:in `block in perform_search'
# /Users/david/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/searchkick-5.3.0/lib/searchkick/multi_search.rb:21:in `each'
# /Users/david/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/searchkick-5.3.0/lib/searchkick/multi_search.rb:21:in `each_with_index'
# /Users/david/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/searchkick-5.3.0/lib/searchkick/multi_search.rb:21:in `perform_search'
# /Users/david/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/searchkick-5.3.0/lib/searchkick/multi_search.rb:11:in `perform'
# /Users/david/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/searchkick-5.3.0/lib/searchkick.rb:188:in `block in multi_search'
# /Users/david/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/activesupport-7.1.1/lib/active_support/notifications.rb:206:in `block in instrument'
# /Users/david/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/activesupport-7.1.1/lib/active_support/notifications/instrumenter.rb:58:in `instrument'
# /Users/david/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/activesupport-7.1.1/lib/active_support/notifications.rb:206:in `instrument'
# /Users/david/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/searchkick-5.3.0/lib/searchkick.rb:187:in `multi_search'
# ./app/services/search/library.rb:136:in `block in execute_multi_search'
```

Our code that is (sometimes) triggering this exception is `Searchkick.multi_search([...])` where among the search options in `[...]` we include `misspellings: { below: 1, prefix_length: 3 }`. The exception occurs when conducting this search with a search string that contains a sufficiently large number of space-separated words.

As can be see from the above exception and stack trace, the exception is occurring [here](https://github.com/ankane/searchkick/blob/v5.3.0/lib/searchkick/results.rb#L92) (in `total_count`, called by `retry_misspellings?`)  https://github.com/ankane/searchkick/blob/243eb5276baa0b9d0e09e0f5fb6f1f7041e2da5e/lib/searchkick/results.rb#L92 because `response["hits"]` is `nil` but then we try to access `["total"]` on it.

This change avoids that exception by only attempting to compare `total_count` to `@misspellings_below` if `response["error"].blank?` (in which case, presumably, `response["hits"]` should be present (a Hash) when we access `["total"]` on it in `retry_misspellings?`).

### DIFF
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -191,7 +191,7 @@ module Searchkick
     end
 
     def retry_misspellings?(response)
-      @misspellings_below && Results.new(searchkick_klass, response).total_count < @misspellings_below
+      @misspellings_below && response["error"].blank? && Results.new(searchkick_klass, response).total_count < @misspellings_below
     end
 
     private
--- a/lib/searchkick/query.rb
+++ b/lib/searchkick/query.rb
@@ -191,7 +191,7 @@ module Searchkick
     end
 
     def retry_misspellings?(response)
-      @misspellings_below && response["error"].blank? && Results.new(searchkick_klass, response).total_count < @misspellings_below
+      @misspellings_below && response["error"].nil? && Results.new(searchkick_klass, response).total_count < @misspellings_below
     end
 
     private
--- a/test/multi_search_test.rb
+++ b/test/multi_search_test.rb
@@ -34,10 +34,7 @@ class MultiSearchTest < Minitest::Test
   end
 
   def test_misspellings_below_with_errored_query
-    store_names ["Product A"]
-    clause_limit = Searchkick.opensearch? ? 1_024 : 65_536
-    search_string = (["Z"] * (clause_limit + 1)).join(" ")
-    products = Product.search(search_string, misspellings: {below: 1})
+    products = Product.search("abc", order: {bad_column: :asc}, misspellings: {below: 1})
     Searchkick.multi_search([products])
     assert products.error
   end
--- a/test/multi_search_test.rb
+++ b/test/multi_search_test.rb
@@ -33,6 +33,15 @@ class MultiSearchTest < Minitest::Test
     assert_equal ["abc", "abd"], products.map(&:name)
   end
 
+  def test_misspellings_below_with_errored_query
+    store_names ["Product A"]
+    clause_limit = 1_024
+    search_string = (["Z"] * (clause_limit + 1)).join(" ")
+    products = Product.search(search_string, misspellings: {below: 1})
+    Searchkick.multi_search([products])
+    assert products.error
+  end
+
   def test_query_error
     products = Product.search("*", order: {bad_column: :asc})
     Searchkick.multi_search([products])
--- a/test/multi_search_test.rb
+++ b/test/multi_search_test.rb
@@ -34,7 +34,7 @@ class MultiSearchTest < Minitest::Test
   end
 
   def test_misspellings_below_with_errored_query
-    products = Product.search("abc", order: {bad_column: :asc}, misspellings: {below: 1})
+    products = Product.search("abc", order: [:bad_field], misspellings: {below: 1})
     Searchkick.multi_search([products])
     assert products.error
   end
--- a/test/multi_search_test.rb
+++ b/test/multi_search_test.rb
@@ -35,7 +35,7 @@ class MultiSearchTest < Minitest::Test
 
   def test_misspellings_below_with_errored_query
     store_names ["Product A"]
-    clause_limit = defined?(Elasticsearch) ? 65_536 : 1_024
+    clause_limit = Searchkick.opensearch? ? 1_024 : 65_536
     search_string = (["Z"] * (clause_limit + 1)).join(" ")
     products = Product.search(search_string, misspellings: {below: 1})
     Searchkick.multi_search([products])
--- a/test/multi_search_test.rb
+++ b/test/multi_search_test.rb
@@ -35,7 +35,7 @@ class MultiSearchTest < Minitest::Test
 
   def test_misspellings_below_with_errored_query
     store_names ["Product A"]
-    clause_limit = 1_024
+    clause_limit = defined?(Elasticsearch) ? 65_536 : 1_024
     search_string = (["Z"] * (clause_limit + 1)).join(" ")
     products = Product.search(search_string, misspellings: {below: 1})
     Searchkick.multi_search([products])


test/multi_search_test.rb
```diff
@@ -33,6 +33,15 @@ def test_misspellings_below_unmet
     assert_equal ["abc", "abd"], products.map(&:name)
   end
 
+  def test_misspellings_below_with_errored_query
+    store_names ["Product A"]
+    clause_limit = 1_024
+    search_string = (["Z"] * (clause_limit + 1)).join(" ")
```
davidrunger-cl, line 39: This is a little awkward. The idea here is just to construct some sort of query that will not return any results (which this won't, because `Z` doesn't match `Product A`) and which causes Elasticsearch to return an error (which this will, because there are 1,025 space-separated words (each of which becomes a query clause), whereas the Elasticsearch default maximum allowed number of query clauses is 1,024).

I wrote the test this way partially because it was in a situation similar to this that we encountered this exception in production.

test/multi_search_test.rb
```diff
@@ -33,6 +33,15 @@ def test_misspellings_below_unmet
...
+    products = Product.search(search_string, misspellings: {below: 1})
+    Searchkick.multi_search([products])
+    assert products.error
```
davidrunger-cl, line 42: The real goal of this test is not so much to `assert products.error` (though that is good to test), but rather to implicitly assert that `Searchkick.multi_search([products])` does not raise an exception, which it currently will on `master` if this test is executed without the change to `lib/searchkick/query.rb` that is included in this PR.

lib/searchkick/query.rb
```diff
@@ -191,7 +191,7 @@ def handle_response(response)
     end
 
     def retry_misspellings?(response)
-      @misspellings_below && Results.new(searchkick_klass, response).total_count < @misspellings_below
+      @misspellings_below && response["error"].blank? && Results.new(searchkick_klass, response).total_count < @misspellings_below
```
davidrunger-cl, line 194: It might make more sense to check `response["hits"].present?`, rather than `response["error"].blank?`. I think that which to choose is mostly a matter of taste/philosophy, and I defer to yours, of course, @ankane (or maybe there's some other approach altogether that you think would be better to avoid the exception that we were able to surface).

davidrunger-cl: On this topic, should we technically be calling either of the `activesupport` methods `present?` or `blank?`? I noticed that there are already places in the code base where we are calling both of these methods, and that works fine, for now, I think, because I think that `searchkick` must be used with either `activerecord` or `mongoid`, which I think in turn ultimately require `active_support/core_ext/object/blank`. However, I think it's _theoretically_ possible that either of those dependencies might stop requiring `active_support/core_ext/object/blank`, in which case we might start getting exceptions in `searchkick` where we call `present?` or `blank?`.

I think that this theoretical issue could be resolved by adding `activesupport` as an explicit dependency in `searchkick.gemspec`, and then explicitly requiring `active_support/core_ext/object/blank` in `lib/searchkick.rb`.

davidrunger-cl: A simpler alternative (versus adding `activesupport` as a dependency) would be to change this to `response["error"].nil?`. I will make that change now.

lib/searchkick/query.rb
```diff
@@ -191,7 +191,7 @@ def handle_response(response)
     end
 
     def retry_misspellings?(response)
-      @misspellings_below && Results.new(searchkick_klass, response).total_count < @misspellings_below
+      @misspellings_below && response["error"].nil? && Results.new(searchkick_klass, response).total_count < @misspellings_below
```
davidrunger-cl, line 194: It might make more sense to check `response["hits"].is_a?(Hash)`, or something like that, rather than `response["error"].nil?`. I think that which to choose is mostly a matter of taste/philosophy, and I defer to yours, of course, @ankane (or maybe there's some other approach altogether that you think would be better to avoid the exception that we were able to surface).

davidrunger-cl:
@ankane : Thank you for triggering CI to run for this PR, and thank you as well for fixing the issues on `master` with the test suite for Mongoid v8.

I'm sorry that this PR wasn't initially compatible with Elasticsearch. We use OpenSearch, and I had run the test suite locally only against that, and hoped that there wouldn't be any difference with Elasticsearch. However, there is a difference, since Elasticsearch and OpenSearch have a different default limit for the maximum number of clauses permitted in a query (which the test that I am adding depends upon).

I have updated the test to be compatible with Elasticsearch (by checking `Searchkick.opensearch?` to determine how many "words" to include in the search query), and I have merged in your recent updates to `master` to fix the issues with Mongoid 8, and, so, if you trigger CI to run again, then I think/hope that it will pass this time.

davidrunger-cl:
When trying to construct an erroring query by exceeding the clause limit (as I had been doing before), one complication was that there is a different default clause limit for Elasticsearch vs OpenSearch, and another issue is that these queries with a large number of clauses were quite slow. Thus, in a best case, the test suite was slowed down significantly. In a worst case, the slow query was enough to cause a timeout error, causing the test to fail/flake.

Therefore, I have updated the test to construct a query in a much simpler way (following the example of other test(s) already in the file) by attempting to order the results by a nonexistent field. This avoids both of the above issues, while still reproducing the relevant exception when run against `master` and verifying that the change in this branch fixes/avoids that error.

ankane:
Great, thanks @davidrunger-cl!

Issue #1663, opened 2024-01-04
`scope_results` outputs incorrect warnings


nholden:
**Describe the bug**
Searchkick outputs a "WARNING: Records in search index do not exist in database" for records that exist in the database when they have been filtered out by the `scope_results` option.

**To reproduce**
Use this code to reproduce when possible:

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"

  gem "activerecord", require: "active_record"
  gem "activejob", require: "active_job"
  gem "sqlite3"
  gem "searchkick", git: "https://github.com/ankane/searchkick.git"
  gem "elasticsearch"
end

puts "Searchkick version: #{Searchkick::VERSION}"
puts "Server version: #{Searchkick.server_version}"

ActiveRecord::Base.establish_connection adapter: "sqlite3", database: ":memory:"
ActiveJob::Base.queue_adapter = :inline

ActiveRecord::Schema.define do
  create_table :products do |t|
    t.string :name
    t.boolean :private
  end
end

class Product < ActiveRecord::Base
  searchkick

  scope :not_private, -> { where(private: false) }
end

Product.reindex
Product.create!(name: "Test", private: true)
Product.search_index.refresh
results = Product.search("test", fields: [:name], scope_results: -> (r) { r.not_private })
puts results
```

In the output, you will see `[searchkick] WARNING: Records in search index do not exist in database: Product 1` even though the record exists in the database.

ankane:
Responded in #1664.

Issue #1570, opened 2022-06-24
Correct pagination metadata with `scope_results`


panteo:
**The problem**

When we apply ActiveRecord `where` filters through `scope_results`, it will throw off pagination.

For pagination to work correctly, we need to use Searchkick's `where` option, but this leads to code duplication and consequently to a bad development and maintenance experience.

**Proposed solution**

I propose this solution, but I don't know all the code base really well and so I don't know if there are edge cases breaking it.

```ruby
# lib/searchkick/results.rb

[...]

    def total_count
      if options[:total_entries]
        options[:total_entries]
      elsif @total_entries    # <------------------------------
        @total_entries        # <------------------------------
      elsif response["hits"]["total"].is_a?(Hash)
        response["hits"]["total"]["value"]
      else
        response["hits"]["total"]
      end
    end
    alias_method :total_entries, :total_count

[...]

    def results_query(records, hits)
      records = Searchkick.scope(records)

      ids = hits.map { |hit| hit["_id"] }
      if options[:includes] || options[:model_includes]
        included_relations = []
        combine_includes(included_relations, options[:includes])
        combine_includes(included_relations, options[:model_includes][records]) if options[:model_includes]

        records = records.includes(included_relations)
      end

      if options[:scope_results]
        records = options[:scope_results].call(records)
        @total_entries = [total_count, records.size].min    # <------------------------------
      end

      Searchkick.load_records(records, ids)
    end

[...]
```

If it wasn't a correct solution, I hope it can inspire you to find a better one.

ankane:
Hey @panteo, check out #1088. While I don't recommend it, if you need to filter in Active Record, you'll need to do pagination there:

```ruby
ids = Product.search("book", select: [:id], load: false).map(&:id)
products = Product.where(id: ids).active.page(1).per(50)
```

Instead, I recommend adding the data you need to filter on to `search_data` and using the `where` option.

PR #1664, opened 2024-01-04
Skip non-existant record warning when using `scope_results`

nholden:
Here's a naive first attempt at addressing https://github.com/ankane/searchkick/issues/1663, where records that have been filtered out by the `scope_results` are raising warnings. It skips missing records warnings whenever `search` is called with the `scope_results` option.

I took this approach because it silences the incorrect warnings without introducing any performance overhead. At first glance, I couldn't think of a way to check if records were missing from the database without performing an additional query for unscoped results.

The downside with this approach is that we'll no longer warnings for hits missing database records when using `scope_results`. This feels worth it to silence incorrect warnings, but I'm definitely open to other approaches here!

### DIFF
--- a/lib/searchkick/results.rb
+++ b/lib/searchkick/results.rb
@@ -269,7 +269,7 @@ module Searchkick
                   # not ideal to return different types
                   # but this situation shouldn't be common
                   model: models.size == 1 ? models.first : models
-                }
+                } unless options[:scope_results]
               end
               result
             end
--- a/test/query_test.rb
+++ b/test/query_test.rb
@@ -84,7 +84,7 @@ class QueryTest < Minitest::Test
     skip unless activerecord?
 
     store_names ["Product A", "Product B"]
-    assert_warns "Records in search index do not exist in database" do
+    refute_warns do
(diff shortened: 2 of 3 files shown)


ankane:
Hey @nholden, thanks for the PR. `scope_results` isn't really meant to be used for filtering, so I don't want to make any changes for it, but to get rid of the warnings, you can either:

1. Use `silence_warnings` from Active Support
2. Use `load: false, select: [:id]` to get the ids, then query Active Record manually (like https://github.com/ankane/searchkick/issues/1570#issuecomment-1166659533)

nholden:
Thanks, @ankane! 🙏

This makes me wonder if I'm using Searchkick incorrectly. In my app today, we have a whole bunch of records and any given user only has access to a subset of them. We index all the records and use `scope_results` on searches to only return the records the current user has access to. Does it sound like there's something fundamentally off with how we're indexing or searching records?

ankane:
Typically, you'll want to do filtering in Elasticsearch, but for more complex / critical cases like authorization, one of the approaches above should be fine.

PR #1668, opened 2024-01-27
fix: remove not exist method

JohnnyKei:
remove `records` delegation.

It does not exist, then it cause a error.
```
NoMethodError: undefined method `records' for #<Searchkick::Relation
```


ankane:
Thanks @JohnnyKei!

PR #1681, opened 2024-06-24
Use external version control

OuYangJinTing:
This PR integrates the external version control function of Elasticsearch with Searchkick, avoiding the problem of incorrect document data in Elasticsearch caused by concurrency during asynchronous synchronization of Searchkick data. For details, you can refer to [this article about Elasticsearch](https://www.elastic.co/blog/elasticsearch-versioning-support).

Note: Because the [Elasticsearch document update operation](https://github.com/elastic/elasticsearch/blob/v7.17.22/server/src/main/java/org/elasticsearch/rest/action/document/RestUpdateAction.java#L77-#L85) does not support version control, when using [Partial Reindexing](#partial-reindexing)
, version control will be ignored.

_PS: English is not my native language; please excuse typing errors._

### DIFF
--- a/README.md
+++ b/README.md
@@ -665,6 +665,35 @@ class Product < ApplicationRecord
 end
 ```
 
+### Versioning
+
+[Elasticsearch Versioning Support](https://www.elastic.co/blog/elasticsearch-versioning-support). Searchkick can enable external version control for Elasticsearch by setting the version_type option. e.g.
+
+```ruby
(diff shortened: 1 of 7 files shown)


ankane:
Hi @OuYangJinTing, thanks for the PR. However, I don't think it's common enough to add the complexity (also, using the time from the server for the external version will have issues with clock skew).

PR #1690, opened 2024-08-07
Improve setup instructions in readme: Add step on setting up credentials

olieidel:
Hey, thanks so much for your work on Searchkick, it's awesome!

I'm currently in the process of using it for one of my projects and I got stuck while setting it up because the readme doesn't mention how to pass your Elasticsearch / Opensearch connection parameters to Searchkick. Took me some reading of the code (specifically `attr_accessor :client_options` in the `Searchkick` class) to understand how to pass them to Searchkick upon Rails initialization.

I added the steps to the readme. If this was somehow obvious and these steps are not helpful, feel free to close!

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,32 @@ gem "opensearch-ruby" # select one
 
 The latest version works with Elasticsearch 7 and 8 and OpenSearch 1 and 2. For Elasticsearch 6, use version 4.6.3 and [this readme](https://github.com/ankane/searchkick/blob/v4.6.3/README.md).
 
+Store your Elasticsearch credentials somewhere, e.g. in the rails credentials storage or in environment variables. If you'd like do use the rails credentails storage, run:
+
+```shell
+EDITOR=nano rails credentials:edit
+```
+
+The credentials file could look like this:
+
+``` yaml
+elasticsearch:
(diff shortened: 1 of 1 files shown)


ankane:
Hi @olieidel, thanks for the PR. There's already a [deployment section](https://github.com/ankane/searchkick?tab=readme-ov-file#deployment) that shows how to set the URL and client options.

olieidel:
Duh, didn't find that, thanks!

That being said, the current deployment section is still missing the TLS stuff for the self-hosted option - starting with Elasticsearch 8 (I think?), TLS encryption is enabled by default, so, at minimum, users would have to either a) copy over the CA certificate (see [step 7 in the install docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html)), or b) pass the CA fingerprint to the client options hash as in my PR (which personally I find simpler).

That being said.. this might only apply to local setups and/or might also be out of scope for the readme here. Just saying that it's a missing step for people setting it up locally for development, and people might get stuck at this step when setting up Searchkick. That being said, totally your call whether to include it or not! 🙂

ankane:
Thanks @olieidel, will think on this.

PR #1695, opened 2024-10-05
Use deep_merge for `model_options`

cumet04:
This PR adds support for nested parameters to `model_options`. For example,
```ruby
Searchkick.model_options = {
  settings: {
    number_of_shards: 1,
    number_of_replicas: 0,
  },
}
```

```ruby
  searchkick settings: {
    analysis: {
      tokenizer: {
        kuromoji_with_dictionary: {...},
      },
      analyzer: { ... },
    },
  }
end
```
with code above, number_of_shards/replicas are ignored previously, but these are considered in this PR.
I guess this behaviour is expected because similar feature client_options is deep_merged.

Note: I haven't written any tests for this PR since there aren't any existing tests for model_options or client_options. Should I add some tests?

### DIFF
--- a/lib/searchkick/model.rb
+++ b/lib/searchkick/model.rb
@@ -1,7 +1,7 @@
 module Searchkick
   module Model
     def searchkick(**options)
-      options = Searchkick.model_options.merge(options)
+      options = Searchkick.model_options.deep_merge(options)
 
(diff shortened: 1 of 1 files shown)


ankane:
Hi @cumet04, thanks for the PR! It looks good, but want to wait until 6.0 to merge since it changes the current behavior.

cumet04:
Thank you for considering this! I'm looking forward to 6.0.

PR #1699, opened 2024-12-09
Fix module name typo in Railtie

natematykiewicz:
I just had a typo in my code:

```ruby
Searckick.search(
```

The error I received was surprising:
```
undefined method `search' for Searckick:Module
```

Turns out, you made the same typo that I did.


ankane:
Nice catch, thanks @natematykiewicz!

PR #1704, opened 2025-01-24
Ensure hits order is respected after model records are loaded when using sort

aovertus:
Hi there, @ankane 

We have an issue on our project while using search with `sort` options then `kaminari` to paginate. 

While the query is properly built and return the proper result in ES, since we load the ActiveRecord using where, the records are not returned in proper order which breaks the purpose of the sort option. 

Let me know if we are missing anything


aovertus:
@fmichaut-diff, @fabioalbini

aovertus:
Hi @ankane this was actually an issue on our side with the indexes, unrelated to the gem itself, wrong solution 😅

PR #1706, opened 2025-03-03
Add support for multiple methods in partial indexing

Yasoob01:
This PR enhances Searchkick's partial indexing functionality by allowing multiple methods to be indexed simultaneously. Previously, only one method could be used at a time, requiring app-level workarounds to merge data from different sources (e.g., client data and buyer data).

We implemented this enhancement in our own project by patching Searchkick’s bulk reindexing, and the results have been positive. Given its usefulness, we believe it’s worth contributing to the main repository.

Changes:
✅ Modified bulk reindexing to support multiple indexing methods.
✅ Ensures backward compatibility with existing single-method indexing.

Let me know if any changes or improvements are needed!

### DIFF
--- a/lib/searchkick/record_data.rb
+++ b/lib/searchkick/record_data.rb
@@ -49,7 +49,16 @@ module Searchkick
     def search_data(method_name = nil)
       partial_reindex = !method_name.nil?
 
-      source = record.send(method_name || :search_data)
+      if method_name.is_a?(Array)
+        source = {}
+        method_name.uniq.each do |method|
+          res = record.send(method)
+          raise "Method #{method} is not a hash" unless res.is_a?(Hash)
+          source.merge!(res)
+        end
+      else
+        source = record.send(method_name || :search_data)
+      end
 
       # conversions
       index.conversions_fields.each do |conversions_field|
--- a/lib/searchkick/record_indexer.rb
+++ b/lib/searchkick/record_indexer.rb
@@ -31,7 +31,7 @@ module Searchkick
           Searchkick::ReindexV2Job.perform_later(
             record.class.name,
             record.id.to_s,
-            method_name ? method_name.to_s : nil,
+            method_name.is_a?(Array) ? method_name.map(&:to_s) : method_name&.to_s,
             routing: routing,
             index_name: index.name
           )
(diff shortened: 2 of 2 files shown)


ankane:
Hi @Yasoob01, thanks for the PR. You can already add a method to the model that calls other methods, so I don't think this is needed.

Yasoob01:
> Hi @Yasoob01, thanks for the PR. You can already add a method to the model that calls other methods, so I don't think this is needed.


You're absolutely right that we can define a method on the model to call other methods. However, the issue arises when we need dynamic method combinations for different reindexing needs.

Right now, every time we need a new combination of fields, we have to modify the model and deploy the change, which is inefficient.


Consider this common pattern in our system:
```
def search_data
  res = {
    id: id,
    value: value,
    updated_at: updated_at,
  }
  .merge(user_data)
  .merge(crm_lead_data)
  .merge(area_data)
  .merge(zone_data)
  .merge(city_data)
  .merge(client_data)
  .merge(freezable_data)
  .merge(lead_source_data)
end
```

Each method (user_data, crm_lead_data, city_data, etc.) returns a specific dataset that we may or may not need during reindexing.

Now, if we need to reindex only city_data and client_data, we currently have to create a new method that includes only those fields:

```
def search_data_city_and_client
  city_data.merge(client_data)
end
```

If tomorrow we need to reindex only crm_lead_data and zone_data, we need another method:

```
def search_data_crm_lead_and_zone
  crm_lead_data
  .merge(zone_data)
end
```

You can see how this quickly becomes unmanageable, as every combination requires a new method, cluttering the model and requiring a deployment for each change.

### **Solution: Dynamic Method Selection (No Deployment Needed)**
**Example 1: Reindex Only city_data and client_data**
```
Searchkick::BulkReindexJob.perform_later(
  class_name: "Lead",
  record_ids: [10, 11, 12],
  index_name: "leads_index",
  method_name: [:city_data, :client_data]
)
```

**Example 2: Reindex Only crm_lead_data and zone_data**
```
Searchkick::BulkReindexJob.perform_later(
  class_name: "Lead",
  record_ids: [10, 11, 12],
  index_name: "leads_index",
  method_name: [:crm_lead_data, :zone_data]
)
```

Now, we don't need to create a new method for every combination, and we don't need a deployment every time the indexing requirements change!
Would love to hear your thoughts on this approach.

PR #1715, opened 2025-05-12
Cross-cluster searches

lucashungaro:
This Pull Request adds the ability to execute Cross-Cluster Searches (CCS).

[Elasticsearch docs](https://www.elastic.co/docs/solutions/search/cross-cluster-search) | [OpenSearch docs](https://docs.opensearch.org/docs/latest/search-plugins/cross-cluster-search/)

CCS is becoming more and more critical as regional data governance and regulations strengthen. In our product, the data for US companies remains stored at rest in US data centers, while data from European companies is stored in Europe. The same is applied to OpenSearch indices. CCS is paramount to allowing admin tools and customers with global reach to have a unified search experience (i.e., a single search provides ranking, pagination out-of-the-box, as opposed to running separate searches and then combining the results).
 
The implementation involves only minor additions to allow specifying remote indices (e.g., remote_cluster:index_name) in search queries. This aligns well with Searchkick's philosophy of offering full Elasticsearch/OpenSearch power while keeping the developer interface simple, and allows developers who need to use CCS to take full advantage of all the fantastic features of Searchkick.

The CCS functionality is optional and does not affect users who don’t configure remote clusters. It’s a non-breaking enhancement that integrates cleanly into the current Searchkick API and index naming structure.

I'm open to answering questions and/or changing the implementation, tests, and documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1809,6 +1809,28 @@ indices_boost: {Category => 2, Product => 1}
 
 Check out [this great post](https://www.tiagoamaro.com.br/2014/12/11/multi-tenancy-with-searchkick/) on the [Apartment](https://github.com/influitive/apartment) gem. Follow a similar pattern if you use another gem.
 
+## Cross-Cluster Searches
+
+To perform cross-cluster searches ([Elasticsearch](https://www.elastic.co/docs/solutions/search/cross-cluster-search) | [OpenSearch](https://docs.opensearch.org/docs/latest/search-plugins/cross-cluster-search/)), use the `ccs_clusters` option:
+
+```ruby
+Product.search("boots", ccs_clusters: ["remote_cluster"])
+# or
+Searchkick.search("boots", models: [Product, Category], ccs_clusters: ["remote_cluster"])
+```
+
+This will perform a search using both the local index (or indices) plus the one(s) in the clusters listed in the array. Assuming the index for `Product` is called `products_development`:
+
+```ruby
+Product.search("boots", ccs_clusters: ["remote_cluster"]) # will perform a search using the local products_development index + the remote_cluster:products_development index
+```
+
+If you want to perform a search using only the remote cluster indices, use the `ccs_exclude_local` option:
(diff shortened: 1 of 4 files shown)


ankane:
Hi @lucashungaro, thanks for the PR. The `index_name` option should already support this.

lucashungaro:
> Hi @lucashungaro, thanks for the PR. The `index_name` option should already support this.

Hi @ankane. Can you help me understand how this would work?

I know we can use an index name prefix, and that would likely work to search on a single remote cluster, for example:

```
https://server/remote_cluster:products/_search
```

But in case we want to search on multiple indices, including the local one + one or more remote clusters like:

```
https://server/remote_cluster_one:products,remote_cluster_two:products,products/_search
```

We could have a dynamic index name via a block that can return such a string (i.e. `remote_cluster_one:products,remote_cluster_two:products,products`), but would that mean we have less flexibility on when we want to use CCS? The main use case here is: end customers will only search on their local clusters, but admin tools will search on a mix of local and remote clusters as needed, all using the same models. So I want to give the developer a lot of flexibility when executing searches from different points in the codebase. The block that defines the index name would have to be aware of that (which clusters should be included in this specific call) and that felt more complex than adding the options to the query class.

I also didn't want to mess with how the gem controls the index names internally, so I chose to use additional options when searching. If we depend on index name resolution and that changes in the future, it could cause Cross-Cluster Search to break.

ankane:
You can use:

```ruby
index_name = Product.search_index.name
products = Product.search("product", index_name: [index_name, "cluster1:#{index_name}", "cluster2:#{index_name}"])
```

(exclude the first item for remote only)

lucashungaro:
@ankane Hah! Nice. I didn't know we could pass `index_name` when calling the `search` method, and somehow didn't notice it as part of the options when I included the ccs ones! I'll test this approach (should work just fine) and open a PR adding that to the documentation. :)

Issue #1119, opened 2018-04-16
Deprecation warning on Elasticsearch 6.2.3


jbwl:
I'd like to report two deprecation warnings for ES 6.2.3

First:
`[WARN ][o.e.d.a.c.NGramTokenFilterFactory] Deprecated big difference between max_gram and min_gram in NGram Tokenizer,expected difference must be less than or equal to: [1]`

This warning is caused by a difference between max and min > 1. If a higher difference is indeed wanted, there is a new setting to allow it (and get rid of the deprecation warning):

>The index level setting index.max_ngram_diff controls the maximum allowed difference between max_gram and min_gram. 

From: https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-ngram-tokenizer.html

Second:
`[WARN ][o.e.d.i.a.ShingleTokenFilterFactory] Deprecated big difference between maxShingleSize and minShingleSize in Shingle TokenFilter,expected difference must be less than or equal to: [3]`

Similar to the first one, there is a new index level setting to allow differences higher than 3 if needed.

Explained here: https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-shingle-tokenfilter.html

ankane:
Awesome, thanks @jbwl for the detailed report and solutions 💯

jbwl:
Thanks for the quick fix!

PR #1716, opened 2025-06-06
Add Elastic Serverless support

jclusso:
- Add new `serverless?` method to detect if serverless.
- Disable unsupported options
- Modified range queries since it appears serverless currently reports 8.11, but expects the newer `gt`, `gte`, `lt`, and `lte` format.

### DIFF
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -150,6 +150,10 @@ module Searchkick
     Gem::Version.new(server_version.split("-")[0]) < Gem::Version.new(version.split("-")[0])
   end
 
+  def self.serverless?
+    @serverless ||= server_info["version"]["build_flavor"] == "serverless"
+  end
+
   # private
   def self.knn_support?
     if opensearch?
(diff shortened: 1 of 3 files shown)


jclusso:
What if this was an option instead of depending on the client info?

As Elastic serverless gains traction there is going to be more users running into a bunch of errors with Searchkick.

ankane:
Feel free to maintain a fork for users who need this (but changing the settings like this will almost certainly break functionality).

jclusso:
Nevermind, I've figured out what things don't work now. This won't work.

jclusso:
@ankane I was wondering if there is any reason this library needs to use ngram and edge ngram. I noticed back in #1119 someone posted about the same error and I'm wondering if Elastic is phasing the ability to set this so high out.

I've been playing around with using prefix and wildcard queries and I was wondering if you'd support a PR that replaces the use of edge ngram and ngram with this? That would allow this library to support serverless.

Issue #1567, opened 2022-06-15
Configurable HashWrapper


philipbjorge:
**Problem**

We introduced some instrumentation that highlighted that `HashWrapper#initialize` accounts for hundreds of milliseconds on some of our endpoints -- Usually between 15-30% of the total request time.

We exclusively use `load: false`, so we're hitting this everywhere.

**Solution**

I'd like to be able to disable the `HashWrapper` and just keep the plain hash here:
https://github.com/ankane/searchkick/blob/744ef325e75a195128433c18813d1ebae4b7059e/lib/searchkick/results.rb#L294

```
# sample code
[Searchkick.hash_wrapper_class.present? ? Searchkick.hash_wrapper_class.new(result) : result, hit]
```

**Additional context**

Some of the worst issues are related to large JSON documents with embedded GeoJSON (using `select:` to improve this), but even after fixing those hotspots, we'd just love to remove this dependency altogether -- It's really costly in the context of our request lifecycle.

Open to other suggestions to bypass this as well if you don't think it's appropriate for the library.

Cheers!

philipbjorge:
Maybe a solution is to work directly with the `hits`?

https://github.com/ankane/searchkick/blob/744ef325e75a195128433c18813d1ebae4b7059e/lib/searchkick/results.rb#L142

ankane:
Hey @philipbjorge, still have the same thoughts as https://github.com/ankane/searchkick/issues/1275#issuecomment-500506063.

PR #1718, opened 2025-06-25
Allow a custom class to be used in place of HashWrapper

DyegoCosta:
This PR introduces the ability to configure a custom wrapper class for Searchkick results, replacing the default `Searchkick::HashWrapper`.

This allows us to avoid loading full ActiveRecord models while:
- Customize the behavior and structure of search results.
- Make small tweaks to the indexed data before it's used in the application (e.g., computed fields, normalized keys, etc.).

The new wrapper class provides a flexible way to shape search results without compromising performance or requiring model hydration.


ankane:
Hi @DyegoCosta, thanks for the PR, but check out #1567.

PR #1721, opened 2025-09-05
Issue #1597 - Fix unscoped detection to allow search within unscoped blocks.

### LINKED ISSUE #1597: detection of relation vs model doesn't handle unscope
**Describe the bug**
I have an application using Rails 6 and Rails-Admin. Upgrading searchkick to 5.x makes all rails_admin pages fail with "search must be called on model, not relation" (because rails_admin always uses a scope). However, even if I exit the scope by calling `unscoped`, this failure still occurs. I believe that in addition to checking for `current_scope.nil?`, searchkick should also check for `current_scope == {}` 

**To reproduce**
Use this code to reproduce when possible:

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"

  gem "sqlite3", platform: :ruby
  gem "activerecord", "~>6.0", require: "active_record"
  gem "activejob", "~>6.0", require: "active_job"
  gem "searchkick", "~>5.1"
  gem "elasticsearch", ">= 7.5", "< 7.14"
end

puts "Searchkick version: #{Searchkick::VERSION}"
puts "Server version: #{Searchkick.server_version}"

ActiveRecord::Base.establish_connection adapter: "sqlite3", database: ":memory:"
ActiveJob::Base.queue_adapter = :inline

ActiveRecord::Schema.define do
  create_table :examples do |t|
    t.string :name
    t.string :description
  end
end

class Example < ActiveRecord::Base
  searchkick

  scope :with_name, ->() { where("NAME IS NOT NULL") }

  def self.custom_search(query)
    unscoped do
      search(query, fields: [:name]).response
    end
  end
end

Example.reindex
Example.create!(name: "Test", description: "Foo")
Example.create!(name: "Test 2", description: "Bar")
Example.create!( description: "Baz")
Example.search_index.refresh
# this should work
p Example.with_name.custom_search("test")
```

**Additional context**

Changing `relation?` to the following definition make unscoped work:

```
module Searchkick
  def self.relation?(klass)
    if klass.respond_to?(:current_scope)
      !(klass.current_scope.nil? || klass.send(:scope_attributes) == {})
    else
      klass.is_a?(Mongoid::Criteria) || !Mongoid::Threaded.current_scope(klass).nil?
    end
  end
end
```

ammar-khalid-05:
### Summary
This PR fixes an issue where `Product.search` could not be called inside an `unscoped` block.  
Previously, Searchkick raised an error when attempting to run searches within `ActiveRecord#unscoped`,  
due to overly strict detection logic.

### Changes
- Updated `relation?` check in `lib/searchkick.rb` to properly handle `unscoped` blocks.
- Added a test (`test_search_unscoped`) to confirm search works as expected within `unscoped`.

### Example
```ruby
Product.where(name: "Test").unscoped do
  Product.search("*", load: false) # ✅ now works without raising error
end
```
Fixes #1597 - Thanks to [@Roguelazer](https://github.com/Roguelazer) for reporting this!

### DIFF
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -352,7 +352,7 @@ module Searchkick
   # it's a bit tricky, but this seems to work
   def self.relation?(klass)
     if klass.respond_to?(:current_scope)
-      !klass.current_scope.nil?
+      !(klass.current_scope.nil? || klass.send(:scope_attributes) == {})
     else
       klass.is_a?(Mongoid::Criteria) || !Mongoid::Threaded.current_scope(klass).nil?
     end
--- a/test/search_test.rb
+++ b/test/search_test.rb
@@ -8,6 +8,16 @@ class SearchTest < Minitest::Test
     assert_equal "search must be called on model, not relation", error.message
   end
 
+  def test_search_unscoped
+    skip unless activerecord?
+    
+    Product.where(name: "Test").unscoped do
+      assert_nothing_raised do
+        Product.search("*", load: false)
+      end
+    end
+  end
+
   def test_body
     store_names ["Dollar Tree"], Store
     assert_equal ["Dollar Tree"], Store.search(body: {query: {match: {name: "dollar"}}}, load: false).map(&:name)


ankane:
Hi @ammar-khalid-05, thanks for the PR, but this breaks the test case directly above it.

PR #1723, opened 2025-09-18
Fix uninitialized constant Searchkick::Mongoid in relation? when Mongoid is not defined

rafayqayyum:
Currently, Searchkick.relation? assumes that if a class does not respond to current_scope, it must be a Mongoid relation. This causes a NameError: uninitialized constant Searchkick::Mongoid in applications that do not use Mongoid.

This PR updates the method to only reference Mongoid::Criteria and Mongoid::Threaded if Mongoid is actually defined.


ankane:
Hi @rafayqayyum, thanks for the PR, but Searchkick only supports Active Record and Mongoid.

PR #1734, opened 2026-01-14
Add support for custom search_document_id with record loading

### LINKED ISSUE #1732: When adding a search_document_id, it is impossible to find a record in the database
**First**
Search existing issues to see if it’s been discussed.
None could be find

**Is your feature request related to a problem? Please describe.**
Because we want to prevent issues when restoring a database backup and keeping the same searchkick index. We want to include a tenant id into the `search_document_id`. Currently, when using search with the load option, this will result in problems. Since searchkick can't find the accompanying record. 

**Describe the solution you'd like**
Add an optional way to find the record in the case the search_id is overridden. Such as a "find_by_search_document_id" method or similar.

**Additional context**
Add any other context.

khasinski:
When a model defines a `custom search_document_id`, searching with `load: true` would fail to find records because Searchkick used the custom ID to query the database primary key.

This adds support for models to define `find_by_search_document_ids` class method to handle the mapping from custom search document IDs back to database records.

Example:
```ruby
    searchkick

    def search_document_id
      "tenant_#{tenant_id}_#{id}"
    end

    def self.find_by_search_document_ids(search_ids)
      ids = search_ids.map { |sid| sid.split("_").last }
      where(id: ids)
    end
  end
```

Fixes #1732

### DIFF
--- a/lib/searchkick/results.rb
+++ b/lib/searchkick/results.rb
@@ -248,7 +248,7 @@ module Searchkick
           grouped_hits.each do |index, index_hits|
             results[index] = {}
             index_models[index].each do |model|
-              results[index].merge!(results_query(model, index_hits).to_a.index_by { |r| r.id.to_s })
+              results[index].merge!(results_query(model, index_hits).to_a.index_by { |r| record_search_id(r) })
             end
           end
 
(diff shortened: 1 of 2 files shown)


sdhull:
We build weird search document ids, so this (or something like it) would be appealing. Considering that there _is_ already a way to customize the search document ids (eg, override `search_document_id`), it makes sense conceptually/from symmetry standpoint to be able to customize the loading of records from search document ids as well.

I'm trying to think of how to make this more symmetrical. Something like `to_search_document_id`/`from_search_document_id` makes sense to me conceptually, but I don't love the idea of deprecating `#search_document_id` instance method just for aesthetics

Anyway, regardless—I like this PR and @ankane you should merge it (or something similar!)

PR #1739, opened 2026-02-12
`Searchkick::Relation#aggs` supports positional and keyword args

james-reading:
Fixes https://github.com/ankane/searchkick/issues/1738


ankane:
Hi @james-reading, thanks for reporting and the PR! Not sure how I missed that. Pushed a small tweak in the commit above for backwards compatibility with the `aggs({store_id: {}})` case.

PR #1746, opened 2026-05-07
Add batch_size: option to Searchkick.callbacks(:bulk)

manoelneto:
## Motivation

When indexing large numbers of records in a `:bulk` callbacks block, a single `_bulk` request can exhaust connection pools or exceed request size limits (OpenSearch/Elasticsearch returns 413). The current workaround is manual `each_slice` outside the block.

## Changes

New `batch_size:` keyword argument on `Searchkick.callbacks(:bulk)` splits accumulated items into bounded chunks, each sent as a separate `_bulk` request **at the end of the block** (not mid-block).

```ruby
Searchkick.callbacks(:bulk, batch_size: 1000) do
  Product.find_each(&:update_fields)
end
```

Key behaviors:

- Items accumulate during the block (no mid-block flush)
- At block end, `Indexer#perform(batch_size:)` slices items internally; each slice sent as own `_bulk` request
- Single `ActiveSupport::Notifications` instrument event wraps all batched HTTP calls (count reflects total items)
- Exception inside block → nothing flushed (clean failure, no partial state)
- Default (no `batch_size:`) → single flush as before, unchanged
- `Searchkick.bulk_batch_size=` setter added, mirroring `callbacks_value=` pattern
- Thread-local state; nested blocks restore outer `batch_size` on exit

## Tests

- All items indexed across batches
- Remainder (items < batch_size) flushed at block end
- Nesting restores outer `batch_size`
- Exception leaves nothing indexed
- N HTTP calls made when batch_size set (5 items / batch_size 2 → 3 `bulk` calls)

## Naming

Used `batch_size:` to match the existing convention in `RelationIndexer`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -651,6 +651,14 @@ Searchkick.callbacks(:bulk) do
 end
 ```
 
+To avoid large `_bulk` requests when processing many records, use `batch_size:` to flush automatically after every N items.
+
+```ruby
+Searchkick.callbacks(:bulk, batch_size: 1000) do
+  Product.find_each(&:update_fields)
+end
+```
+
 Or temporarily skip updates.
 
 ```ruby
--- a/lib/searchkick.rb
+++ b/lib/searchkick.rb
@@ -214,11 +214,13 @@ module Searchkick
   end
 
   # message is private
-  def self.callbacks(value = nil, message: nil)
+  def self.callbacks(value = nil, batch_size: nil, message: nil)
     if block_given?
       previous_value = callbacks_value
+      previous_batch_size = bulk_batch_size
       begin
         self.callbacks_value = value
+        Thread.current[:searchkick_bulk_batch_size] = batch_size if value == :bulk
         result = yield
         if callbacks_value == :bulk && indexer.queued_items.any?
           event = {}
@@ -235,10 +237,15 @@ module Searchkick
         result
       ensure
         self.callbacks_value = previous_value
+        Thread.current[:searchkick_bulk_batch_size] = previous_batch_size
       end
     else
       self.callbacks_value = value
     end
+  end
+
+  def self.bulk_batch_size
+    Thread.current[:searchkick_bulk_batch_size]
   end
 
   def self.aws_credentials=(creds)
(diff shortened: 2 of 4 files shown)


ankane:
Hi @manoelneto, thanks for the PR, but I don't see a strong reason to add an option for this (it's easy enough to batch manually).

manoelneto:
Hi @ankane, thanks for taking a look. Let me share the concrete use case — it’s harder to batch manually than it may initially seem.

We generate `DocumentChunk` records from `Articles`, `Pages`, and `Users`, then index them into OpenSearch. Two constraints make a built-in `batch_size` option for `Searchkick.callbacks(:bulk)` especially valuable:

1. Chunk counts per record vary wildly and aren’t known up front. A single record may produce anywhere from 1 to 1000+ chunks depending on content length. We process records in outer batches (~10), so the resulting chunk count per flush fluctuates significantly.

2. Our OpenSearch index uses an ingest pipeline with an ML inference processor. That pipeline runs per-document on the OpenSearch side and is both slow and sensitive to payload size. Large bulk requests trigger “request too large” and timeout errors, and we can’t realistically tune the pipeline around arbitrary payload sizes.

The “just batch manually” workaround isn’t quite a one-liner for this shape.

Without a `batch_size:` option, we have to:

* Disable Searchkick callbacks entirely (`Searchkick.callbacks(false)`)
* Track every chunk created and every chunk soft-deleted during processing
* Re-slice that combined list ourselves
* Call `.reindex` per slice inside `Searchkick.callbacks(:bulk)`

That’s effectively what we ended up implementing:

```ruby
Searchkick.callbacks(false) do
  created_chunks = process_batch(batch) # creates + soft-deletes chunks
  reindex_chunks_in_batches(created_chunks) if created_chunks.any?
end

def reindex_chunks_in_batches(chunks)
  chunks.each_slice(REINDEX_BATCH_SIZE) do |chunk_batch|
    Searchkick.callbacks(:bulk) do
      chunk_batch.each(&:reindex)
    end
  end
end
```

This works, but it pushes the responsibility onto every caller dealing with:

* unpredictable fan-out
* downstream payload limits
* throughput-sensitive indexing pipelines

A `batch_size:` option on `Searchkick.callbacks(:bulk)` would let Searchkick flush internally in controlled chunks, while callers could continue using the natural callback flow instead of bypassing it and re-implementing batched flushing manually.

Happy to adjust the API shape or add tests/docs if there’s a direction you’d prefer.

